PR #30, opened 2015-10-28
Add travis-ci testing

euank:
This adds a `.travis.yml` file and some misc things needed for it to work there.

One thing this added which shouldn't necessarily be there forever is the environment variables in the `make test` target. It would be better to not have such a dependency at all.

This also splits `generate` into a target which is not run every time in the interest of reducing dependencies when running tests and also in making the code-test-build cycle quicker.

Tested on my fork https://travis-ci.org/euank/amazon-ecs-cli

r? @samuelkarp @uttarasridhar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: go
+sudo: false
+os:
+    - linux
+    - osx
+go:
+    - 1.4
+    - 1.5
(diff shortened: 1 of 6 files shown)


samuelkarp:
:+1:

uttarasridhar:
:+1:

PR #35, opened 2015-11-02
Correcting credential logic.

nrdlngr:
The current credential logic seems wrong for what it is trying to do. The logic should check for profile, and if profile is empty, then fail if one of access key or secret key is also empty.
With the current logic, my system had the env vars set for access key and secret key, and then the `ecs-cli configure` command failed with the `--profile` flag.


lox:
Looks like your PR includes way more than you meant to @nrdlngr.

PR #39, opened 2015-11-09
Fix release builds appearing dirty

samuelkarp:
With this change, release builds (using the release script) no longer appear dirty:

```
$ ./ecs-cli -v
ecs-cli version 0.1.0 (1bff41b)
```

r? @uttarasridhar @euank


euank:
LGTM, though it would be nice to make the clean marker more accurate (both here and in the Agent where this is sourced from).

uttarasridhar:
:+1

PR #50, opened 2015-12-15
Change ordering of AWS Credential resolution for the ecs-cli

juanrhenals:
This commit streamlines the logic for resolving AWS credentials.
The ecs-cli now attempts to resolve AWS credentials in the following
order: ENV variables, ECS Profile, AWS Profile and EC2 Instance role.
This commit also removes some unneccessary logic in the ecs-cli
configure command.

Fixes #35 

@samuelkarp @aaithal @euank @uttarasridhar

### DIFF
--- a/ecs-cli/modules/config/config.go
+++ b/ecs-cli/modules/config/config.go
@@ -26,7 +26,11 @@ import (
 
 // TODO: This needs a better home.
 
-const RegionFlag = "region"
+const (
+	RegionFlag = "region"
+	// This time.Minute value comes from the SDK defaults package
+	eC2RoleProviderExpiryWindow = 5 * time.Minute
+)
 
 // CliConfig is the top level struct used to map to the ini config.
 type CliConfig struct {
@@ -61,6 +65,8 @@ func (cfg *CliConfig) ToServiceConfig() (*aws.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// This is just a fail-fast check to ensure that valid credentials are available before returning to the caller.
 	if creds.AccessKeyID == "" {
 		return nil, fmt.Errorf("Error getting valid credentials")
 	}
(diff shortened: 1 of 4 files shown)


aaithal:
lgtm

ecs-cli/modules/config/config.go
```diff
@@ -26,7 +26,10 @@ import (
 
 // TODO: This needs a better home.
 
-const RegionFlag = "region"
+const (
+	RegionFlag                  = "region"
+	EC2RoleProviderExpiryWindow = 5 * time.Minute
```
euank: This can be private, and it might make sense to reference where this time comes from (looks like it was the SDK defaults package).

ecs-cli/modules/config/config.go
```diff
@@ -49,18 +52,26 @@ func NewCliConfig(cluster string) *CliConfig {
 }
 
 // ToServiceConfig creates an aws Config object from the CliConfig object.
-func (cfg *CliConfig) ToServiceConfig() (*aws.Config, error) {
+func (cfg *CliConfig) ToServiceConfig(providers ...credentials.Provider) (*aws.Config, error) {
```
euank: Since there aren't actually any callers that pass in `providers` right now (tests or otherwise), I'm in favour of not having the argument and then someone can add it if/when tests that need that mocking are added.

ecs-cli/modules/config/config.go
```diff
@@ -74,43 +85,34 @@ func (cfg *CliConfig) ToServiceConfig() (*aws.Config, error) {
...
-	// The 'default' profile credential provider is next. Add if only if profile is not
-	// set in ecs config.
-	if cfg.AwsProfile == "" {
-		credentialProviders = append(credentialProviders, &credentials.SharedCredentialsProvider{
+		},
+		&credentials.SharedCredentialsProvider{
+			Filename: "",
+			Profile:  cfg.AwsProfile,
+		},
+		&credentials.SharedCredentialsProvider{
```
euank: If cfg.AwsProfile isn't set, you'll already have added one with `Profile: ""` right above, which gets the commented behaviour of "profile not specified, use default".

What this does do is make it so a set, but non-existent, profile will not error but will instead become the default profile. Is this intended?

ecs-cli/modules/config/config_test.go
```diff
@@ -14,86 +14,106 @@
...
-	if len(credentialProviders) != 3 {
-		t.Fatal("Unexpected number of credential providers in the chain: ", len(credentialProviders))
+	ecsConfig.Region = region
+
+	os.Setenv("AWS_ACCESS_KEY_ID", envAwsAccessKey)
+	os.Setenv("AWS_SECRET_ACCESS_KEY", envAwsSecretKey)
+
+	// Clear env variables as they persist past the individual test boundary
+	defer func() {
+		os.Unsetenv("AWS_ACCESS_KEY_ID")
```
euank: If our tests are free of other environment dependencies, `os.Clearenv()` will be more readable.

ecs-cli/modules/config/config_test.go
```diff
@@ -14,86 +14,106 @@
...
+
+func TestToServiceConfigWhenRegionIsNotSpecified(t *testing.T) {
+	ecsConfig := NewCliConfig(clusterName)
+
+	_, err := ecsConfig.ToServiceConfig()
+	if err == nil {
+		t.Error("There should always be an error when region is not specified in the ecsConfig.")
+	}
+}
+func TestToServiceConfigWhenNoCredentialsAreAvailable(t *testing.T) {
```
euank: I think we should skip this test until we either mock out the providers or, at least, mock out the ec2 role provider to an extent (e.g. inject a mock http client). Maybe this is a use-case for overriding the provider chain.

I say that because this test will fail on an EC2 instance with a role attached or could take 5 seconds to pass due to the http timeout.

juanrhenals: I'll add a skip for TestToServiceConfigWhenNoCredentialsAreAvailable(..)

ecs-cli/modules/config/config_test.go
juanrhenals: As discussed, I'll leave this as is.

ecs-cli/modules/config/config.go
juanrhenals: I have removed the extra SharedCredentialsProvider such that the remaining SharedCredentialsProvider either attempts to use the specified Profile OR the default provider depending on whether the cfg.AwsProfile is non-empty.

ecs-cli/modules/config/config.go
juanrhenals: Removed.

ecs-cli/modules/config/config.go
juanrhenals: Privatized.

euank:
:+1:

PR #58, opened 2016-01-07
Update AMIs from 2015.09.b to 2015.09.c.

### LINKED ISSUE #56: EC2 Container Registry Support
Currently, the AMI in the ecs-cli is 2015.03.b, which is the ECS Agent 1.1.0
https://github.com/aws/amazon-ecs-cli/blob/cff682c389b5d863a44f8e38972fd039f4517f9b/ecs-cli/modules/config/ami/ami.go#L29

ECR requires that the ECS Agent 1.7.0 or greater, so we should update for the latest AMI of 2015.03.g.
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html

Upgrading this will allow ECR images to be used with clusters created by the ecs-cli.

skion:
Got these from http://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html

This should fix #56.

### DIFF
--- a/ecs-cli/modules/config/ami/ami.go
+++ b/ecs-cli/modules/config/ami/ami.go
@@ -28,12 +28,14 @@ type staticAmiIds struct {
 
 func NewStaticAmiIds() ECSAmiIds {
 	regionToId := make(map[string]string)
-	// amzn-ami-2015.09.b-amazon-ecs-optimized AMIs
+	// amzn-ami-2015.09.c-amazon-ecs-optimized AMIs
 	regionToId["us-east-1"] = "ami-6ff4bd05"
 	regionToId["us-west-1"] = "ami-46cda526"
 	regionToId["us-west-2"] = "ami-313d2150"
 	regionToId["eu-west-1"] = "ami-8073d3f3"
+	regionToId["eu-central-1"] = "ami-60627e0c"
 	regionToId["ap-northeast-1"] = "ami-6ca38b02"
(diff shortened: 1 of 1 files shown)


ecs-cli/modules/config/ami/ami.go
```diff
@@ -29,12 +29,12 @@ type staticAmiIds struct {
 func NewStaticAmiIds() ECSAmiIds {
 	regionToId := make(map[string]string)
 	// amzn-ami-2015.09.b-amazon-ecs-optimized AMIs
```
samuelkarp: Can you update the comment here?

samuelkarp:
Thanks for doing this!  I believe it should be safe to add the two new regions as well.

frewsxcv:
Just tried to use the ECS Registry with ecs-cli and was sad to see it wasn't natively supported. But glad to see it was addressed in this PR. Thanks @skion! :confetti_ball:

frewsxcv:
The fail on Travis CI seems unrelated, no?

euank:
Reran the failed one; it looked related to a [transient Github issue](https://status.github.com/).

:+1: from me.

aaithal:
lgtm as well

samuelkarp:
:+1:

juanrhenals:
:+1:

sentientmonkey:
Thanks again, Pieter! :beer:

skion:
Thank you for the fast turnover!

PR #62, opened 2016-01-13
Allow creation of non-essential containers by setting 'restart' parameter to 'no'

rthrfrd:
At the moment it is impossible to create an ECS task definition from a compose file that uses data-only containers. This is because all containers are created as 'Essential', so when the data containers exit (as they should), the whole task is stopped.

Using the `restart` field fits within the compose.yml spec, does not affect existing behaviour, and also makes sense semantically:
- With `restart` set to `'no'` (thus `Essential` being `false`) the Task will not be killed, so an ECS service _won't_ reschedule a new instance of the Task as the existing one will continue to run.
- With `restart` set to any other value (thus `Essential` defaulting to `true`) the Task will be killed, so an ECS service _will_ reschedule a new instance of the Task as the existing one will not continue to run.

P.S. I don't write Go, apologies for any mistakes.

Thanks,

Alex

### DIFF
--- a/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
+++ b/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
@@ -144,7 +144,7 @@ func convertToContainerDef(inputCfg *libcompose.ServiceConfig,
 	outputContDef.EntryPoint = aws.StringSlice(inputCfg.Entrypoint.Slice())
 	outputContDef.Environment = environment
 	if inputCfg.Restart == "no" {
-		outputContDef.Essential = aws.Bool(true)
+		outputContDef.Essential = aws.Bool(false)
 	}
 	outputContDef.ExtraHosts = extraHosts
 	if inputCfg.Hostname != "" {
--- a/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
+++ b/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
@@ -143,7 +143,7 @@ func convertToContainerDef(inputCfg *libcompose.ServiceConfig,
 	outputContDef.DockerSecurityOptions = aws.StringSlice(inputCfg.SecurityOpt)
 	outputContDef.EntryPoint = aws.StringSlice(inputCfg.Entrypoint.Slice())
 	outputContDef.Environment = environment
-	if inputCfg.restart == "no" {
+	if inputCfg.Restart == "no" {
 		outputContDef.Essential = false
 	}
 	outputContDef.ExtraHosts = extraHosts
--- a/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
+++ b/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
@@ -143,6 +143,9 @@ func convertToContainerDef(inputCfg *libcompose.ServiceConfig,
 	outputContDef.DockerSecurityOptions = aws.StringSlice(inputCfg.SecurityOpt)
 	outputContDef.EntryPoint = aws.StringSlice(inputCfg.Entrypoint.Slice())
 	outputContDef.Environment = environment
+	if inputCfg.restart == "no" {
+		outputContDef.Essential = false
+	}
 	outputContDef.ExtraHosts = extraHosts
 	if inputCfg.Hostname != "" {
 		outputContDef.Hostname = aws.String(inputCfg.Hostname)
--- a/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
+++ b/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
@@ -144,7 +144,7 @@ func convertToContainerDef(inputCfg *libcompose.ServiceConfig,
 	outputContDef.EntryPoint = aws.StringSlice(inputCfg.Entrypoint.Slice())
 	outputContDef.Environment = environment
 	if inputCfg.Restart == "no" {
-		outputContDef.Essential = false
+		outputContDef.Essential = aws.Bool(true)
 	}
 	outputContDef.ExtraHosts = extraHosts
 	if inputCfg.Hostname != "" {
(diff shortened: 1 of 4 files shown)


ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
```diff
@@ -143,6 +143,9 @@ func convertToContainerDef(inputCfg *libcompose.ServiceConfig,
 	outputContDef.DockerSecurityOptions = aws.StringSlice(inputCfg.SecurityOpt)
 	outputContDef.EntryPoint = aws.StringSlice(inputCfg.Entrypoint.Slice())
 	outputContDef.Environment = environment
+	if inputCfg.Restart == "no" {
```
uttarasridhar: Thanks a lot for the pull request.

Compose options that ECS supports are whitelisted in [this file](https://github.com/aws/amazon-ecs-cli/blob/master/ecs-cli/modules/compose/ecs/utils/convert_compose_yml.go#L29). You will have to explicitly add "restart" there.
Can you also add a unit test for this change?

rthrfrd:
@uttarasridhar Thanks, I've had a go!

tallavi:
+1

xaduha:
@uttarasridhar any progress on this? Is there any other way to make it non essential currently?

rthrfrd:
@xaduha My ugly hack while testing was to make the container ping itself to keep it running:
```yml
    command:
        - "ping"
        - "127.0.0.1"
        - "-q"
```

xaduha:
@rthrfrd Doesn't help me, unfortunately.

bryceray1121:
It has been a month? Any updates on getting this merged?

rthrfrd:
@samuelkarp @uttarasridhar Is this project still maintained? Thanks

jesucarr:
+1

sanglt:
+1

leandrocr:
C'mon guys +1

NinjaPT:
+1 please

NinjaPT:
hi @uttarasridhar any update on getting this merged?

rthrfrd:
I would not expect this project to be maintained. Strategically speaking, controlling the orchestration of containers is probably the best way for AWS/Google/Docker etc to create any kind of vendor lock-in. This is to counter the fact that the containers themselves are portable and thus commoditise the infrastructure those services sell. That's also why Docker is recently pushing to integrate orchestration more directly into the docker engine via swarm. Interoperability between the ECS/Kubernetes/swarm etc formats does not further that goal.

ambrons:
@rthrfrd yeah this is a bit depressing, but true.  A co-worker of mine showed me this the other day as well:
http://thenewstack.io/docker-fork-talk-split-now-table/

We've been considering switching to docker 1.12 and using swarm with it's orchestration and / or also using marathon with Mesosphere.

rthrfrd:
@ambrons We're eventually going with Kubernetes based on maturity, features and size of the ecosystem outside of Google. I'm sceptical of swarm - while I love `docker` for allowing us to containerise our infrastructure (which has brought huge benefits in spite of the constant bugs), I don't trust Docker Inc to continue making good decisions for us. They are under pressure to justify their funding valuations, so forks appearing is understandable. I also think Mesos is a great option however not such a good fit for our existing setup/team.

uttarasridhar:
Thinking more about this PR, docker `restart` option doesn't quite map 1:1 to Amazon ECS `essential` parameter. We support Docker `restart` functionality using [ECS Services](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_services.html), which maintains a desired count of ECS tasks by launching new tasks for failed/stopped ones.  However, `essential` parameter on the ECS Task Definition just defines how tasks get to the stopped state (if an essential container fails/stops).
 
To explain further, Docker `restart` option [states](https://docs.docker.com/engine/reference/run/#/restart-policies-restart):
> Using the --restart flag on Docker run you can specify a restart policy for how a container should or should not be restarted on exit
 
Whereas, Amazon ECS `essential` parameter [states](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_environment):
> If the essential parameter of a container is marked as true, and that container fails or stops for any reason, all other containers that are part of the task are stopped. If the essential parameter of a container is marked as false, then its failure does not affect the rest of the containers in a task. If this parameter is omitted, a container is assumed to be essential.
 
The way Amazon ECS handles `restart` is:
* if you want restart = `no`, then you could start the task using `ecs-cli compose up` command, which doesn't restart if the task dies.
* if you want restart, then you can start the task using `ecs-cli compose service up` comand, which creates an ECS service, and always ensures that you have a task up and running.
However, the task‘s stop or kill criteria is defined by the `essential` field.
 
I understand that ecs-cli doesn't have the capability today to mark a container as 'non-essential'. To address this, we could expose our own custom compose option or add this feature to Docker compose.
Instead of overloading `restart` for this purpose (which could lead to confusing and unexpected task/container lifecycle behavior), you could use ECS Service to launch new tasks to always guarantee a specific number of running tasks.

For more information around Amazon ECS as an orchestration platform, you can read this related issue explaining the benefits: https://github.com/aws/amazon-ecs-agent/issues/466

ambrons:
@uttarasridhar perhaps there's some synergy to be had with this request for changes to docker-compose.
https://github.com/docker/compose/issues/1896

I can understand not wanting to strong arm a feature in.  Perhaps the larger issue is the lack of control on how the docker-compose file is transformed into a ecs task. Perhaps there could be a _side-car_ hook for allow people to set configuration of how cloudformation and or the task creation should be have.  Think `.ebextensions`

yinshiua:
@ambrons Thanks for your effort of giving this a shot, but docker restart option is not exactly the same as Amazon ECS essential parameter. Base on @uttarasridhar's comment previously, a better way to support this is to figure out a way to expose our own custom compose option. I've opened a new issue to track this https://github.com/aws/amazon-ecs-cli/issues/267. Feel free to give us any suggestions.

PR #66, opened 2016-01-18
Update AMIs from 2015.09.c to 2015.09.d

npars:
Sourced from http://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html


uttarasridhar:
:+1:

Issue #64, opened 2016-01-14
Certain instance types not supported


dtest:
Using ecs-cli to setup a cluster, I get an error for unsupported instance type for 'm4.xlarge' or 'm4.2xlarge':

```bash
$ ecs-cli --version
ecs-cli version 0.1.0 (*UNKNOWN)
$ ecs-cli up   --keypair *snip*   --capability-iam   --size 1   --vpc *snip*  --subnets *snip*  --azs us-east-1a   --instance-type m4.xlarge
INFO[0000] Created cluster                               cluster=opsviz-dev
ERRO[0000] Error executing 'up': ValidationError: Parameter EcsInstanceType failed to satisfy constraint: must be a valid EC2 instance type.
	status code: 400, request id: a90c31b7-badf-11e5-a984-97970441ee29
```

While I couldn't find what ImageID the cloudformation template uses, according to the documentation on setting up an ECS instance, ECS would use the following ami for us-east-1 region: ami-2b3b6041 (http://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html)

If I use that AMI to create the instance itself, it indicates support for the m4.* instance classes. 

c4.2xlarge works fine via ecs-cli

Are various instance types not supported? If so, is this documented somewhere?

Also, sidenote: Is the ECS cloudformation template available somewhere?

dtest:
Actually, I found the cloudformation template. It's defining these supported instance types:

```bash
"EcsInstanceType": {
      "Type": "String",
      "Description": "ECS EC2 instance type",
      "Default": "t2.micro",
      "AllowedValues": [
        "t2.micro",
        "t2.small",
        "t2.medium",
        "t2.large",
        "m3.medium",
        "m3.large",
        "m3.xlarge",
        "m3.2xlarge",
        "c4.large",
        "c4.xlarge",
        "c4.2xlarge",
        "c4.4xlarge",
        "c4.8xlarge",
        "c3.large",
        "c3.xlarge",
        "c3.2xlarge",
        "c3.4xlarge",
        "c3.8xlarge",
        "r3.large",
        "r3.xlarge",
        "r3.2xlarge",
        "r3.4xlarge",
        "r3.8xlarge",
        "i2.xlarge",
        "i2.2xlarge",
        "i2.4xlarge",
        "i2.8xlarge"
      ]
```

Any particular reason m4.* is not supported?

samuelkarp:
[...]

PR #67, opened 2016-01-19
Add support for t2.nano, g2 and d2 instance types

uttarasridhar:
Addresses issue #64. Got the updated instance types list from https://aws.amazon.com/ec2/instance-types/


aaithal:
:shipit:

PR #81, opened 2016-02-13
ECS-CLI Amazon ELB Support

abdollar:
ECS-CLI Amazon ELB Support

Not sure why amazon ELB support is not provided by the amazon ecs-cli tool. Added a pull request to provide a simple configuration to get this started using the ecs-cli tool.
- Uses the cloud formation template to construct an ELB associated with your vpc
- Adds an extra parameter for the SSL certificate called certificate-arn.

```
certificate-arn = arn:aws:acm:us-east-1:000000000000:certificate/00000000-0000-0000-0000-000000000000
```
- Assumes the ECS instance uses port 80 and the ELB uses 443 and terminates SSL - this should ideally be optional and somehow configured but seems to be a good default.
- Opens Egress to everything for the ELB and only 443 and 80 for the ECS instance as it makes http calls - this again should be configurable.
- Added a cloudformation template that adds extra logging using the fluent logging driver - again should be configurable as not logging your docker container logs seems wrong.Assumes you have added a task called fluentd-task and image for getting docker logs to wherever you need using fluentd
- Assumes you have a role called ecsServiceRole that connects ELB and ECS.

The IAM policy for the ecsServiceRole looks like this

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "elasticloadbalancing:Describe*",
        "elasticloadbalancing:DeregisterInstancesFromLoadBalancer",
        "elasticloadbalancing:RegisterInstancesWithLoadBalancer",
        "ec2:Describe*",
        "ec2:AuthorizeSecurityGroupIngress"
      ],
      "Resource": [
        "*"
      ]
    }
  ]
}
```

This tries to provide an out of the box configuration for connecting amazon ELB and Amazon ECS but should be more configurable so that custom configurations can be easily altered.

This is something that I created in a short time frame that needs extra work to parameterize more, but the default should ideally work for simple configurations.

### DIFF
--- a/ecs-cli/modules/aws/clients/cloudformation/params.go
+++ b/ecs-cli/modules/aws/clients/cloudformation/params.go
@@ -34,6 +34,7 @@ const (
 	ParameterKeyKeyPairName   = "KeyName"
 	ParameterKeyCluster       = "EcsCluster"
 	ParameterKeyAmiId         = "EcsAmiId"
+	ParameterKeyCertificate   = "Certificate"
 )
 
 var ParameterNotFoundError = errors.New("Parameter not found")
--- a/ecs-cli/modules/aws/clients/cloudformation/template.go
+++ b/ecs-cli/modules/aws/clients/cloudformation/template.go
@@ -13,8 +13,29 @@
 
 package cloudformation
 
+import (
+       "io/ioutil"
+       "os"
+       "path/filepath"
+
+       "github.com/Sirupsen/logrus"
+)
+
 func GetTemplate() string {
-	return template
+       templateFile := "cloudformation.tmpl"
+       if envTemplate := os.Getenv("ECS_CLOUDFORMATION_TEMPLATE"); envTemplate != "" {
+               templateFile = envTemplate
+       } else if wd, err := os.Getwd(); err != nil {
+               templateFile = filepath.Join(wd, templateFile)
+       }
+       if _, err := os.Stat(templateFile); err == nil {
+               buffer, err := ioutil.ReadFile(templateFile)
+               if err == nil && len(buffer) > 0 {
+                       logrus.Debugf("using cloudformation template file: %s", templateFile)
+                       return string(buffer)
+               }
+       }
+       return template
 }
 
 // TODO: Improvements:
@@ -131,6 +152,11 @@ var template = `
     "EcsCluster" : {
       "Type" : "String",
       "Description" : "ECS Cluster Name",
+      "Default" : "default"
+    },
+    "Certificate": {
+      "Type" : "String",
+      "Description" : "Certificate ARN for SSL certificates for ELB",
       "Default" : "default"
     }
   },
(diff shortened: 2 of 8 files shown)


davemkirk:
I'm interested in seeing ELB support added to ecs-cli but I see a potential issue with this implementation. 

The launch configuration user data now assumes the cluster host will come with `yum`. Line 529 in the change to `ecs-cli/modules/aws/clients/cloudformation/template.go`

+`"yum install -y aws-cli jq\n",`

Won't this be a problem where an ami without yum is passed in from the `--image-id` flag?

abdollar:
@davemkirk there's more than one issue - it assumes a lot - the whole cloud formation snippet makes a lot of assumptions - it currently hard codes a task name. It should be parameterized somehow, not sure how - maybe a file?
Also, the health check is hard coded as well and needs to be parameterized.
Also, the egress and ingress rules make assumptions as well that need to be parameterized.
I would describe this as a lot of work in progress.

abdollar:
triggering travis build by closing and re-opening

toddpi314:
This would be helpful, however... https://github.com/aws/amazon-ecs-cli/issues/90

TakaraShinya:
This would be helpful.
I'm waiting for ...

kaihendry:
Support for ELBs in `ecs-cli` is sorely needed. I haven't run this code, but here are my comments after comparing it to my shell scripts:

* security certificate can easily be added after the fact [on the console](https://ap-southeast-1.console.aws.amazon.com/ec2/v2/home?region=ap-southeast-1#LoadBalancers:) so I don't think an initial parameter is warranted

My scripts here:
1. https://github.com/kaihendry/count/blob/master/create-task.sh - assumption **project name** (basename dirname of CWD)?
2. https://github.com/kaihendry/count/blob/master/create-load-balancer.sh - ask to choose existing security group instead of creating a new one. Also do not make assumptions with subnet ( you have PubSubnetAz1?)
3. https://github.com/kaihendry/count/blob/master/create-service.sh - I assume 9000 for my container port, so I'm assuming your `containerDef.PortMappings[0].ContainerPort` works better =)

* You probably want to remove "github.com/Sirupsen/logrus"

You focus a bit on policies and I am not sure why. I seem to be able to escape them [in my cluster setup procedure](https://github.com/kaihendry/count).


* service-name "ecscompose-service-$PROJECT_NAME"
* task-definition "ecscompose-$PROJECT_NAME"
* loadBalancerName=$CLUSTER_NAME which is the same as $PROJECT_NAME (not EcsCluster as you have it)

Hope this helps and if I have the time & you think my suggestions are ok, I'd happily contribute.

abdollar:
@kaihendry I don't use this command very often and I tend to rely on the aws cli as you do in your scripts as well as a cloud formation template. 

To try and answer your questions. In terms of the certificate - I understand you can use the console, though some companies have policies where developers have no console access and so providing that out of the box helps. Ideally, this updates route53 as well - assuming thats your DNS provider.

This PR makes a lot of assumptions with regard to the network setup. Specifically, the egress and ingress rules as well as subnets. That should be changed or configurable and I believe they have some changes to use an existing VPC rather than creating a new one.

Feel free to contribute - this is really not my repository at all and I don't have any connection with amazon aws and this cli tool at all - this is just a PR I added to show how this worked given some changes with the ELB. I don't think this is a good tool to use for production use - its good for development to get you started to simulate docker compose. It would be good to use the amazon ALB rather than the ELB as that supports multiple containers on a single instance.

yinshiua:
There is a PR https://github.com/aws/amazon-ecs-cli/pull/211 which allows you to specify an existing ELB and ALB.

Closing this PR because we would want to support ALB instead of ELB. ALB allows you to define routing rules based on content across multiple containers running on one ore more EC2 instances which better fit our use case.

PR #100, opened 2016-04-08
Update the AMI ids to Amazon ECS-Optimized Amazon Linux AMI 2016.03.a

uttarasridhar:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html


samuelkarp:
:+1:

Issue #125, opened 2016-07-12
project-name parameter broken in v0.4.0


esetnik:
The `project-name` parameter is dropping special characters. I use underscores in my project-name and the current release broke my deploy process because all the underscores are dropped.

`ecs-cli compose --project-name backend-service-dev --file docker-compose.yml service up`

```
INFO[0000] Using ECS task definition                     TaskDefinition=backendservicedevelop:4
INFO[0000] Created an ECS service                        service=backendservicedevelop taskDefinition=backendservicedevelop:4
INFO[0000] Updated ECS service successfully              desiredCount=1 serviceName=backendservicedevelop
INFO[0000] Describe ECS Service status                   desiredCount=1 runningCount=0 serviceName=backendservicedevelop
INFO[0015] ECS Service has reached a stable state        desiredCount=1 runningCount=1 serviceName=backendservicedevelop
```

PR #127, opened 2016-07-12
Preserve compose project name as user has defined it.

uttarasridhar:
closes aws/amazon-ecs-cli#125.

CONTEXT
---
Recently, ecs-cli was migrated to use libcompose for parsing the compose project (including project name and compose yml files). 
Libcompose normalizes the project name by removing '-' and other non alpha-numeric values from the name as seen in the commit https://github.com/docker/libcompose/commit/ca237071578647c6fef35f6d5a9e424bb6deca03, libcompose. This behavior is not documented anywhere in [Docker Compose project_name doc](https://docs.docker.com/compose/reference/envvars/#compose-project-name).
As a result, the project name parsing wasn't backward compatible with ecs-cli. This commit addresses that concern and resets the project name to what the user set (after libcompose parses it).

TEST
---
$ make generate

$ make test
PASS

$ make docker-build
Built ecs-cli for linux
Built ecs-cli for darwin

$ make build
. ./scripts/shared_env && ./scripts/build_binary.sh ./bin/local
Built ecs-cli

$ ./bin/local/ecs-cli compose --project-name proj-name up
INFO[0000] Using ECS task definition                     TaskDefinition=ecscompose-proj-name:1

---

### DIFF
--- a/ecs-cli/modules/compose/cli/ecs/app/command.go
+++ b/ecs-cli/modules/compose/cli/ecs/app/command.go
@@ -86,7 +86,7 @@ func commonComposeFlags() []cli.Flag {
 			EnvVar: "COMPOSE_FILE",
 		},
 		cli.StringFlag{
-			Name:   projectNameFlag + ",p",
+			Name:   ecscompose.ProjectNameFlag + ",p",
 			Usage:  "Specifies the project name to use. Defaults to the current directory name.",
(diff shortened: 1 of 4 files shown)


aaithal:
:shipit:

samuelkarp:
:+1:

PR #133, opened 2016-07-15
Parse different volumes_from format supported by compose.

uttarasridhar:
# Issue
Compose supports volumes_from with access mode. See [here](https://docs.docker.com/compose/compose-file/#/volumes-from) for details.

If compose file had the option:
```
volumes_from:
- webserver:rw
```
ecs would return an error: volumesFrom.sourceContainer contains invalid characters

# Resolution
Add parsing logic to support the different formats
```
Compose file format v2:
volumes_from:
 - service_name
 - service_name:ro
 - container:container_name
 - container:container_name:rw

Compose file format v1:
volumes_from:
- service_name
- service_name:ro
- container_name
- container_name:rw
```

# Test
$ make test
PASS

$ make build

```
compose yml excerpt:
  volumes_from:
    - webserver:rw

```
$ ./bin/local/ecs-cli compose up
INFO[0000] Using ECS task definition                     TaskDefinition=ecscompose-amazon-ecs-cli:24

```
task definition excerpt:
                "volumesFrom": [
                    {
                        "readOnly": false, 
                        "sourceContainer": "webserver"
                    }
                ], 
```

### DIFF
--- a/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
+++ b/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
@@ -35,6 +35,7 @@ const (
 	// access mode with which the volume is mounted
 	readOnlyVolumeAccessMode  = "ro"
 	readWriteVolumeAccessMode = "rw"
+	volumeFromContainerKey    = "container"
 )
 
 // supported fields/options from compose YAML file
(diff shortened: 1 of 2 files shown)


aaithal:
:shipit:

samuelkarp:
:+1:

PR #147, opened 2016-08-18
Extend README.md - download specific version

nvtkaszpir:
Extend README with info how to download specific version from the S3, because sometimes you do not want to download the `latest` version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,26 @@ Download the binary archive for your platform, decompress the archive, and
 install the binary on your `$PATH`. You can use the provided `md5` hash to 
 verify the integrity of your download.
 
+
+### Latest version
 * Linux: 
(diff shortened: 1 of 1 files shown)


witsoej:
Edit on this change (revised text below):

+Using the URLs above, replace `latest` with the desired tag, for example `v0.4.1`. After downloading, remember to rename the binary file to "ecs-cli".

kaihendry:
I don't see the point of downloading an out of date binary. Old binaries for example have old hard coded AMIs which can have security or reliability issues. If anything there needs to be a guidelines how to trigger `curl -o /usr/local/bin/ecs-cli https://s3.amazonaws.com/amazon-ecs-cli/ecs-cli-linux-amd64-latest` when there is a new release!

nvtkaszpir:
@kaihendry we had an issue that latest binary was producing errors and thus was unusable. That's why this commit was added.

yinshiua:
Resolved conflict and pushed. (Commit https://github.com/aws/amazon-ecs-cli/commit/b476ef6ee7b87737d6429b5e98a48cfa4386d614)

Issue #148, opened 2016-08-22
complex commands when using `ecs-cli compose run [container] [command]`


pgeez:
I cannot get the following command to run on ECS:

`ecs-cli compose run django_command python,manage.py,migrate`

because it is mis-parsed as the following

![image](https://cloud.githubusercontent.com/assets/1369475/17843547/1f7dc1a4-67e4-11e6-9e84-2c5bc8cca72f.png)

It should be parsed as an array, in accordance with the behavior of the ECS web interface.

Is there **any** way to run commands via `ecs-cli` without defining brand new tasks that can have multiple arguments?

pgeez:
does something like this work?

see: https://github.com/aws/amazon-ecs-cli/blob/master/ecs-cli/modules/aws/clients/ecs/client.go#L405

replace
```
Command: aws.StringSlice([]string{command}),
```

with 
```
Command: aws.StringSlice(strings.Split(command, ",")),
```

jsh2134:
you could try putting that command into a shell script and calling that

$ cat migrate.sh
```
#/bin/sh
python manage.py migrate
```

ECS Command: migrate.sh

yinshiua:
The PR has been merged. Resolving...

PR #150, opened 2016-08-26
Partial support for multi-argument commands with compose run

pgeez:
Addresses Issue #148 

Now, a user can pass commands with multiple arguments as a comma-delimited list into the ECS-CLI as:

`ecs-cli compose run <container> cmd0,cmd1,cmd2`

NOTE: subcommands with embedded commas are not yet supported, though this could be added in the future through the use of `encoding/csv` with an agreed-upon command-line delimiter escape sequence.

### DIFF
--- a/ecs-cli/modules/aws/clients/ecs/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/client.go
@@ -17,6 +17,7 @@ import (
 	"crypto/md5"
 	"errors"
 	"fmt"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/aws/amazon-ecs-cli/ecs-cli/modules/aws/clients"
@@ -402,7 +403,7 @@ func (client *ecsClient) RunTaskWithOverrides(taskDefinition, startedBy string, 
 	for cont, command := range overrides {
 		contOverride := &ecs.ContainerOverride{
 			Name:    aws.String(cont),
-			Command: aws.StringSlice([]string{command}),
+			Command: aws.StringSlice(strings.Split(command, ",")),
 		}
 		commandOverrides = append(commandOverrides, contOverride)
 	}
--- a/ecs-cli/modules/aws/clients/ecs/client_test.go
+++ b/ecs-cli/modules/aws/clients/ecs/client_test.go
@@ -18,6 +18,8 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/aws/amazon-ecs-cli/ecs-cli/modules/aws/clients"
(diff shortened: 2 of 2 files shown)


akirakoyasu:
ping @yinshiua
I long this feature, too. It's critical for `run`'s capability.
How do you think?

yinshiua:
@akirakoyasu I will look into this in the next few days.

pgeez:
Thank you! This feature would really, really help a lot. The change I
implement is a simple one-liner to split input on whitespace; that did the
trick for the basic use cases.

pgeez:
*split input on whitespace --> split input on commas

assuming the commands were entered as comma-delimited. proper CSV escaping
is definitely required for this approach, which I didn't do. i.e. how do
you handle multipart commands that have commas in them?

yinshiua:
@pgeez Does the last suggestion (Putting the command in shell script) in issue https://github.com/aws/amazon-ecs-cli/issues/148 fit your use case?

Rather than going with the comma delimited approach, we should use go-shlex to help analyze the input command, and it would be a much better user experience.

pgeez:
Great idea! Admittedly, my solution was a short-term hack to enable this
functionality.

pgeez:
And, no. I don't see why an additional shell script is required. Having
behavior that replicates docker-compose run cmd arg0 arg1 ... would be way,
way better and more maintainable.

jsh2134:
@yinshiua I would also love to see this feature enabled in order to maintain the same behavior as docker-compose

yinshiua:
> Now, a user can pass commands with multiple arguments as a comma-delimited list into the ECS-CLI as:
`ecs-cli compose run <container> cmd0,cmd1,cmd2`

@pgeez I understand that we want the behavior to be similar to docker-compose and this is why I am opposed to merge this PR. Your current implementation is comma-delimited but docker-compose is space separated. In addition, even if we change to space separated , how would you parse a command like `sh -c echo $HOME` to `["sh", "-c", "echo $HOME"]` (there is a space between `echo` and `$HOME`)?

pgeez:
Yes, I'm not a Go developer and I don't know the correct way to do this in
Go which is why I hacked in comma-delimiting.

I agree that feature-parity requires supporting the format:

compose run container cmd0 cmd1 cmd2 ...

as you pointed out. Does Go's arg parser not support this natively? (For
reference, I know that Python's native argparse module supports narg
commands natively). How does Docker Compose handle it?

pgeez:
@yinshiua <https://github.com/yinshiua> To be 100% clear, my intention was
to get something that works so that we -- the ECS-CLI community -- could
get AWS eyes on the problem. I don't care if you close my pull request; in
fact, given it's incompatibility with the Docker Compose syntax, I'd
suggest closing the PR and implementing something that works correctly.
This feature is massively important.

yinshiua:
@pgeez Go does not natively support this, we will have to use the library I mentioned above. Docker compose uses python which as you mention is supported natively.

We will prioritize this in our next release. Thanks!

pgeez:
Awesome! Looking forward to it.

akirakoyasu:
@pgeez @yinshiua Thanks a lot, I understand what we want.

If breaking changes are acceptable, I can implement this.
Please see above PR; #299

pgeez:
Yes. Changes that break my PR are completely fine; I don't care at all.

yinshiua:
Proposal in PR https://github.com/aws/amazon-ecs-cli/pull/229

Issue #63, opened 2016-01-13
Running latest ecs-cli fails with "ERRO[0000] Error executing 'up': ParameterValue and UsePreviousValue not set for parameter key 'EcsCluster'"


ryanwalls:
Run this:
```
ecs-cli up --keypair my-key-pair --capability-iam --size 2 --instance-type t2.medium
```
and I get this:
```
ERRO[0000] Error executing 'up': ParameterValue and UsePreviousValue not set for parameter key 'EcsCluster'
```

uttarasridhar:
You will need to configure the cluster using the command : ecs-cli configure --cluster <name> before you execute any other commands (ecs-cli up, down, compose, ...)
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_CLI_Configuration.html

ryanwalls:
Thanks.  Might want to add that as Step 1 in your tutorial. Or make a note of it.  http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_CLI_tutorial.html

If the docs are on github, let me know the repo and I'll create a PR for the updated documentation.

lodotek:
I agree, this threw me off as well...

martinrehfeld:
+1 ;-)

jyaworski:
I just ran into this. It's been five months since this was reported.

denis-yuen:
:+1:

benhtn:
Also missing from http://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli.html

PR #155, opened 2016-09-07
Validate cluster name before running Cloudformation Stack

yinshiua:
Better validation message when running ecs-cli up without configuring a cluster name. 
https://github.com/aws/amazon-ecs-cli/issues/63


samuelkarp:
:+1:

PR #23, opened 2015-10-14
Don't create Host key in Volume if no hostPath is defined

mcanevet:
Fixes #11 

Please note that this is my very first go patch so it may have to be refactored.
I couldn't find a way to declare `ecsVolume` out of the `if else` block, and there is probably a better way to initialize the `ecs.Volume` structure without duplicating code.
Comments very welcome :-)

PR #201, opened 2017-02-10
Fixes #11: Error attempting to ecs-cli compose up

### LINKED ISSUE #11: Error attempting to ecs-cli compose up
Created a cluster running on two instances like specified in the examples.

This is my compose file (`docker-compose-prod.yml`):

``` yaml
web:
    image: nathanleclaire/laraveldocker_web
    links:
     - db
     - redis
     - beanstalkd
    ports:
     - "80:80"
    volumes:
     - /var/lib/lightning/mysql:/var/lib/lightning/mysql
db:
    image: mysql
    environment:
        MYSQL_ROOT_PASSWORD: password
    volumes:
     - /var/lib/mysql
redis:
    image: redis
beanstalkd:
    image: nathanleclaire/laraveldocker_beanstalkd
```

When I run:

``` console
$ ecs-cli compose -f docker-compose-prod.yml up
ERRO[0000] Error registering task definition             error=ClientException: host.sourcePath should not be blank.
        status code: 400, request id: b7a2c30b-6e0f-11e5-a4a1-6b2efea4c01d family=0x820c5e110
ERRO[0000] Create task definition failed                 error=ClientException: host.sourcePath should not be blank.
        status code: 400, request id: b7a2c30b-6e0f-11e5-a4a1-6b2efea4c01d
FATA[0000] ClientException: host.sourcePath should not be blank.
        status code: 400, request id: b7a2c30b-6e0f-11e5-a4a1-6b2efea4c01d
```

Seems that `sourcePath` is not getting set as it should be in the task definition.

tiffanyfj:
Fixes #11 

-Cherry-picked PR #23 (thanks @mcanevet)
-The above PR made it so it wouldn't fail, but the same volume was shared amongst all container paths without a host path, so I added this
-Updated test
-Tested it with `ecs-cli compose up` on random compose file I mushed together:

```
data-container:
  image: redis:latest
  volumes:
    - /tmp/stuff 
    - /tmp/bar:/foo/baz
    - /tmp/ponies
    - /tmp/foo:/tmp/bar
    - /tmp/foo:/tmp/baz
  mem_limit: 100M

data-container2:
  image: redis:latest
  volumes:
    - /tmp/hi:/tmp/hello
    - /tmp/foo:/tmp/foo
    - /tmp/unicorns
  mem_limit: 100M
```

### DIFF
--- a/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
+++ b/ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
@@ -48,6 +48,11 @@ var supportedComposeYamlOptions = []string{
 
 var supportedComposeYamlOptionsMap = getSupportedComposeYamlOptionsMap()
 
+type volumes struct {
+	volumeWithHost  map[string]string
+	volumeEmptyHost []string
+}
+
 func getSupportedComposeYamlOptionsMap() map[string]bool {
 	optionsMap := make(map[string]bool)
 	for _, value := range supportedComposeYamlOptions {
@@ -67,7 +72,9 @@ func ConvertToTaskDefinition(taskDefinitionName string, context *project.Context
 	logUnsupportedConfigFields(context.Project)
 
 	containerDefinitions := []*ecs.ContainerDefinition{}
-	volumes := make(map[string]string) // map with key:=hostSourcePath value:=VolumeName
+	volumes := &volumes{
+		volumeWithHost: make(map[string]string), // map with key:=hostSourcePath value:=VolumeName
+	}
 
 	for _, name := range serviceConfigs.Keys() {
 		serviceConfig, ok := serviceConfigs.Get(name)
@@ -83,6 +90,7 @@ func ConvertToTaskDefinition(taskDefinitionName string, context *project.Context
 		}
 		containerDefinitions = append(containerDefinitions, containerDef)
 	}
+
 	taskDefinition := &ecs.TaskDefinition{
 		Family:               aws.String(taskDefinitionName),
 		ContainerDefinitions: containerDefinitions,
@@ -156,7 +164,7 @@ func isZero(v reflect.Value) bool {
 // convertToContainerDef transforms each service in the compose yml
 // to an equivalent container definition
 func convertToContainerDef(context *project.Context, inputCfg *config.ServiceConfig,
-	volumes map[string]string, outputContDef *ecs.ContainerDefinition) error {
+	volumes *volumes, outputContDef *ecs.ContainerDefinition) error {
 	// setting memory
 	var mem int64
 	if inputCfg.MemLimit != 0 {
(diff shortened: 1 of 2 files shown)


ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
```diff
@@ -48,6 +48,11 @@ var supportedComposeYamlOptions = []string{
 
 var supportedComposeYamlOptionsMap = getSupportedComposeYamlOptionsMap()
 
+type volumes struct {
+	volumeHasHost   map[string]string
```
aaithal: `volumeHasHost` seems like something that you'd use for a bool. Can you use a better name here?

ecs-cli/modules/compose/ecs/utils/convert_task_definition_test.go
```diff
@@ -28,12 +28,15 @@ import (
...
-	containerPath = "/tmp/cache"
-	hostPath      = "./cache"
+	portNumber     = 8000
+	portMapping    = "8000:8000"
+	containerPath  = "/tmp/cache"
+	containerPath2 = "/tmp/cache2"
+	hostPath       = "./cache"
 )
 
+var emptyHostCtr = 0
```
aaithal: This doesn't need to be a global variable I think.

ecs-cli/modules/compose/ecs/utils/convert_task_definition_test.go
```diff
@@ -387,18 +406,23 @@ func TestConvertToMountPoints(t *testing.T) {
...
+	} else {
+		sourceVolume = volumes.volumeEmptyHost[emptyHostCtr]
+		emptyHostCtr++
+	}
 	if sourceVolume != *output.SourceVolume {
 		t.Errorf("Expected sourceVolume [%s] But was [%s]", sourceVolume, *output.SourceVolume)
 	}
 	if readonly != *output.ReadOnly {
-		t.Errorf("Expected readonly [%s] But was [%s]", readonly, *output.ReadOnly)
+		t.Errorf("Expected readonly [%v] But was [%v]", readonly, *output.ReadOnly)
```
aaithal: Have you tried using `testify/assert` package for these kinds of things? Makes it less verbose and more elegant.

ecs-cli/modules/compose/ecs/utils/convert_task_definition.go
tiffanyfj: Yeah

ecs-cli/modules/compose/ecs/utils/convert_task_definition_test.go
tiffanyfj: I didn't because that would require rewriting the entire test file or just changing the section I was dealing with.

ecs-cli/modules/compose/ecs/utils/convert_task_definition_test.go
tiffanyfj: Will try again. Wherever I had put it before resulted in it overwriting itself every time it was called.

tiffanyfj: Basically I think I would just have to add another parameter to the function being called otherwise. Should I do that instead?

ecs-cli/modules/compose/ecs/utils/convert_task_definition_test.go
```diff
@@ -387,18 +406,23 @@ func TestConvertToMountPoints(t *testing.T) {
...
+	hostPath, containerPath string, readonly bool, emptyHostCtr *int) {
+	sourceVolume := ""
 	if containerPath != *output.ContainerPath {
 		t.Errorf("Expected containerPath [%s] But was [%s]", containerPath, *output.ContainerPath)
 	}
-	sourceVolume := volumes[hostPath]
+	if hostPath != "" {
+		sourceVolume = volumes.volumeWithHost[hostPath]
+	} else {
+		sourceVolume = volumes.volumeEmptyHost[*emptyHostCtr]
```
aaithal: I think that hard-coding this index might be better than incrementing using a pointer. We already hardcode the index for the `mountPointsOut` slice when this is invoked. What do you think about not incrementing this variable, but passing in a hard-coded value from where it's invoked? So it'd be `verifyMountPoint(t *testing.T, output *ecs.MountPoint, volumes *volumes, hostPath, containerPath string, readonly bool, emptyHostIndex int)`. And then you'll invoke it with the appropriate index from `TestConvertToMountPoints()`. Does that make sense?

ecs-cli/modules/compose/ecs/utils/convert_task_definition_test.go
tiffanyfj: Adding testify/assert will be a separate PR since it's not specific to this issue and isn't currently used in the file.

PR #211, opened 2017-02-23
AWS ELB-ALB support

victorcoder:
We want to be able to create a new service and assign it an Application Load Balancer.

Add 5 new params that allows to specify the Load Balancer details in order for the service to be created with an associated ELB/ALB.

Specify the Target Group ARN, Load Balancer Name, Container Name, Container Port and Role to allow the service to register new instances in the ELB/ALB.

* `--target-group-arn`: The full Amazon Resource Name (ARN) of the Elastic Load Balancing target  group associated with a service.
* `--load-balancer-name`: The name of the classic load balancer.
* `--container-name`: The name of the container (as it appears in a container definition) to associate with the load balancer.
* `--container-port`: The port on the container to associate with the load balancer. This port must correspond to a containerPort in the service's task definition. Your container instances must allow ingress traffic on the hostPort of the port mapping.
* `--role`: The name or full Amazon Resource Name (ARN) of the IAM role that allows Amazon ECS to make calls to your load balancer on your behalf. This parameter is required if you are using a load balancer with your service. If you specify the role parameter, you must also specify a load balancer object with the loadBalancers parameter.

If `--target-group-arn` or `--load-balancer-name` is not passed then the current behaviour is maintained and it doesn't try to associate an ELB with the service.

Note: If both `target-goup-arn` and `load-balancer-name` present, `target-group-arn` param will take precedence.

### DIFF
--- a/ecs-cli/modules/aws/clients/ecs/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/client.go
@@ -134,28 +134,19 @@ func (client *ecsClient) DeleteService(serviceName string) error {
 }
 
 func (client *ecsClient) CreateService(serviceName, taskDefName string, loadBalancer *ecs.LoadBalancer, role string, deploymentConfig *ecs.DeploymentConfiguration) error {
-	var err error
-
-	if *loadBalancer.TargetGroupArn != "" {
-		_, err = client.client.CreateService(&ecs.CreateServiceInput{
-			DesiredCount:            aws.Int64(0),            // Required
-			ServiceName:             aws.String(serviceName), // Required
-			TaskDefinition:          aws.String(taskDefName), // Required
-			Cluster:                 aws.String(client.params.Cluster),
-			DeploymentConfiguration: deploymentConfig,
-			LoadBalancers:           []*ecs.LoadBalancer{loadBalancer},
-			Role:                    aws.String(role),
-		})
-	} else {
-		_, err = client.client.CreateService(&ecs.CreateServiceInput{
-			DesiredCount:            aws.Int64(0),            // Required
-			ServiceName:             aws.String(serviceName), // Required
-			TaskDefinition:          aws.String(taskDefName), // Required
-			Cluster:                 aws.String(client.params.Cluster),
-			DeploymentConfiguration: deploymentConfig,
-		})
+	createServiceInput := &ecs.CreateServiceInput{
+		DesiredCount:            aws.Int64(0),            // Required
+		ServiceName:             aws.String(serviceName), // Required
+		TaskDefinition:          aws.String(taskDefName), // Required
+		Cluster:                 aws.String(client.params.Cluster),
+		DeploymentConfiguration: deploymentConfig,
 	}
-	if err != nil {
+	if *loadBalancer.TargetGroupArn != "" {
+		createServiceInput.LoadBalancers = []*ecs.LoadBalancer{loadBalancer}
+		createServiceInput.Role = aws.String(role)
+	}
+
+	if _, err := client.client.CreateService(createServiceInput); err != nil {
 		log.WithFields(log.Fields{
 			"service": serviceName,
 			"error":   err,
--- a/ecs-cli/modules/aws/clients/ecs/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/client.go
@@ -141,7 +141,8 @@ func (client *ecsClient) CreateService(serviceName, taskDefName string, loadBala
 		Cluster:                 aws.String(client.params.Cluster),
 		DeploymentConfiguration: deploymentConfig,
 	}
-	if *loadBalancer.TargetGroupArn != "" {
+
+	if *loadBalancer.TargetGroupArn != "" || *loadBalancer.LoadBalancerName != "" {
 		createServiceInput.LoadBalancers = []*ecs.LoadBalancer{loadBalancer}
 		createServiceInput.Role = aws.String(role)
 	}
--- a/ecs-cli/modules/aws/clients/ecs/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/client.go
@@ -142,7 +142,8 @@ func (client *ecsClient) CreateService(serviceName, taskDefName string, loadBala
 		DeploymentConfiguration: deploymentConfig,
 	}
 
-	if aws.StringValue(loadBalancer.TargetGroupArn) != "" || aws.StringValue(loadBalancer.LoadBalancerName) != "" {
+	if loadBalancer != nil &&
+		(aws.StringValue(loadBalancer.TargetGroupArn) != "" || aws.StringValue(loadBalancer.LoadBalancerName) != "") {
 		createServiceInput.LoadBalancers = []*ecs.LoadBalancer{loadBalancer}
 		createServiceInput.Role = aws.String(role)
 	}
--- a/ecs-cli/modules/aws/clients/ecs/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/client.go
@@ -142,7 +142,7 @@ func (client *ecsClient) CreateService(serviceName, taskDefName string, loadBala
 		DeploymentConfiguration: deploymentConfig,
 	}
 
-	if *loadBalancer.TargetGroupArn != "" || *loadBalancer.LoadBalancerName != "" {
+	if aws.StringValue(loadBalancer.TargetGroupArn) != "" || aws.StringValue(loadBalancer.LoadBalancerName) != "" {
 		createServiceInput.LoadBalancers = []*ecs.LoadBalancer{loadBalancer}
 		createServiceInput.Role = aws.String(role)
 	}
--- a/ecs-cli/modules/aws/clients/ecs/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/client.go
@@ -47,7 +47,7 @@ type ECSClient interface {
 	IsActiveCluster(clusterName string) (bool, error)
 
 	// Service related
-	CreateService(serviceName, taskDefName string, deploymentConfig *ecs.DeploymentConfiguration) error
+	CreateService(serviceName, taskDefName string, loadBalancer *ecs.LoadBalancer, role string, deploymentConfig *ecs.DeploymentConfiguration) error
 	UpdateServiceCount(serviceName string, count int64, deploymentConfig *ecs.DeploymentConfiguration) error
 	UpdateService(serviceName, taskDefinitionName string, count int64, deploymentConfig *ecs.DeploymentConfiguration) error
 	DescribeService(serviceName string) (*ecs.DescribeServicesOutput, error)
@@ -133,14 +133,28 @@ func (client *ecsClient) DeleteService(serviceName string) error {
 	return nil
 }
 
-func (client *ecsClient) CreateService(serviceName, taskDefName string, deploymentConfig *ecs.DeploymentConfiguration) error {
-	_, err := client.client.CreateService(&ecs.CreateServiceInput{
-		DesiredCount:            aws.Int64(0),            // Required
-		ServiceName:             aws.String(serviceName), // Required
-		TaskDefinition:          aws.String(taskDefName), // Required
-		Cluster:                 aws.String(client.params.Cluster),
-		DeploymentConfiguration: deploymentConfig,
-	})
+func (client *ecsClient) CreateService(serviceName, taskDefName string, loadBalancer *ecs.LoadBalancer, role string, deploymentConfig *ecs.DeploymentConfiguration) error {
+	var err error
+
+	if *loadBalancer.TargetGroupArn != "" {
+		_, err = client.client.CreateService(&ecs.CreateServiceInput{
+			DesiredCount:            aws.Int64(0),            // Required
+			ServiceName:             aws.String(serviceName), // Required
+			TaskDefinition:          aws.String(taskDefName), // Required
+			Cluster:                 aws.String(client.params.Cluster),
+			DeploymentConfiguration: deploymentConfig,
+			LoadBalancers:           []*ecs.LoadBalancer{loadBalancer},
+			Role:                    aws.String(role),
+		})
+	} else {
+		_, err = client.client.CreateService(&ecs.CreateServiceInput{
+			DesiredCount:            aws.Int64(0),            // Required
+			ServiceName:             aws.String(serviceName), // Required
+			TaskDefinition:          aws.String(taskDefName), // Required
+			Cluster:                 aws.String(client.params.Cluster),
+			DeploymentConfiguration: deploymentConfig,
+		})
+	}
 	if err != nil {
 		log.WithFields(log.Fields{
 			"service": serviceName,
--- a/ecs-cli/modules/compose/cli/ecs/app/service_command.go
+++ b/ecs-cli/modules/compose/cli/ecs/app/service_command.go
@@ -138,10 +138,10 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
 }
 
 func loadBalancerFlags() []cli.Flag {
-	targetGroupArnUsageString := "[Optional] The full Amazon Resource Name (ARN) of the Elastic Load Balancing target group associated with a service. Defaults to none"
-	containerNameUsageString := "[Optional] Mandatory if target-group-arn is specified. The container name (as it appears in a container definition). Defaults to none"
-	containerPortUsageString := "[Optional] Mandatory if target-group-arn is specified. The container port to access from the load balancer. Defaults to none"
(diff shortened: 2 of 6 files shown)


ecs-cli/modules/aws/clients/ecs/client.go
```diff
@@ -133,14 +133,28 @@ func (client *ecsClient) DeleteService(serviceName string) error {
...
-		DesiredCount:            aws.Int64(0),            // Required
-		ServiceName:             aws.String(serviceName), // Required
-		TaskDefinition:          aws.String(taskDefName), // Required
-		Cluster:                 aws.String(client.params.Cluster),
-		DeploymentConfiguration: deploymentConfig,
-	})
+func (client *ecsClient) CreateService(serviceName, taskDefName string, loadBalancer *ecs.LoadBalancer, role string, deploymentConfig *ecs.DeploymentConfiguration) error {
+	var err error
+
+	if *loadBalancer.TargetGroupArn != "" {
```
samuelkarp: You'll need to check if `loadBalancer` is `nil` before dereferencing it.

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,36 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
 		},
 	}
 }
+
+func loadBalancerFlags(specifyDefaults bool) []cli.Flag {
```
samuelkarp: @nrdlngr Can you review this text?

ecs-cli/modules/aws/clients/ecs/client.go
victorcoder: As per `loadContext` method https://github.com/aws/amazon-ecs-cli/pull/211/files#diff-97b4106c54aad3d402f437209e9fc4e3R78 `loadBalancer` could have empty values but it'll never be `nil`.

ecs-cli/modules/compose/cli/ecs/app/service_command.go
victorcoder: Just FYI, the text is copy pasted from the param description in the AWS API docs.

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,42 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
+	targetGroupArnUsageString := "[Optional] The full Amazon Resource Name (ARN) of the Elastic Load Balancing target group associated with a service."
+	containerNameUsageString := "[Optional] Mandatory if target-group-arn is specified. The container name (as it appears in a container definition)."
+	containerPortUsageString := "[Optional] Mandatory if target-group-arn is specified. The container port to access from the load balancer."
+	loadBalancerNameUsageString := "[Optional] The name of the load balancer."
+	roleUsageString := "[Optional] The name or full Amazon Resource Name (ARN) of the IAM role that allows Amazon ECS to make calls to your load balancer on your behalf. This parameter is required if you are using a load balancer with your service. If you specify the role parameter, you must also specify a target-group-arn, container-name and container-port."
+
+	if specifyDefaults {
```
yinshiua: When will this be false?

Also, I don't think we need to specify the defaults as none. "Optional" already indicates that it will be empty.

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,42 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
+		targetGroupArnUsageString += " Defaults to none"
+		containerNameUsageString += " Defaults to none"
+		containerPortUsageString += " Defaults to none"
+		loadBalancerNameUsageString += " Defaults to none"
+		roleUsageString += " Defaults to ecsServiceRole"
```
yinshiua: Does it actually defaults to `ecsServiceRole`?

Also, to specify defaults use `Value` in the cli.StringFlag

victorcoder: I took license here and used the role name used in AWS docs http://docs.aws.amazon.com/AmazonECS/latest/developerguide/service_IAM_role.html on creating an IAM role for ECS services.

ecs-cli/modules/compose/cli/ecs/app/service_command.go
victorcoder: You're right, it should never be called with false. Removed

ecs-cli/modules/compose/cli/ecs/app/service_command.go
yinshiua: @victorcoder I'm seeing 2 issues here.

1. You are specifying in the command help text that the role defaults to `ecsServiceRole`, but in your actual implementation of createService, you're not setting the default to `ecsServiceRole`
2. By setting the default role, we are assuming everyone has created the role `ecsServiceRole` which is not always the case, we should be creating the role for them if the role does not exist.

I would leave out the defaults in the command help text for now, and add it back once we have the implementations. Hope that makes sense!

ecs-cli/modules/compose/ecs/flags.go
```diff
@@ -19,4 +19,10 @@ const (
 	DeploymentMaxPercentFlag                = "deployment-max-percent"
 	DeploymentMinHealthyPercentDefaultValue = 100
 	DeploymentMinHealthyPercentFlag         = "deployment-min-healthy-percent"
+	TargetGroupArnFlag                      = "target-group-arn"
+	ContainerNameFlag                       = "container-name"
+	ContainerPortFlag                       = "container-port"
+	LoadBalancerNameFlag                    = "load-balancer-name"
+	RoleDefaultValue                        = "ecsServiceRole"
```
yinshiua: Should also remove the defaults for roles.

ecs-cli/modules/compose/ecs/service.go
```diff
@@ -65,6 +67,27 @@ func (s *Service) LoadContext() error {
...
+		return err
+	}
+	loadBalancerName := s.Context().CLIContext.String(LoadBalancerNameFlag)
+
+	s.loadBalancer = &ecs.LoadBalancer{
+		ContainerName: aws.String(containerName),
+		ContainerPort: containerPort,
+	}
+
+	if targetGroupArn != "" {
```
yinshiua: It would be nice to educate the customers that `loadBalancerName` and `targetGroupArn` by throwing an error when both are specified instead of picking one for them.

ecs-cli/modules/aws/clients/ecs/client.go
```diff
@@ -133,15 +133,21 @@ func (client *ecsClient) DeleteService(serviceName string) error {
...
 		DesiredCount:            aws.Int64(0),            // Required
 		ServiceName:             aws.String(serviceName), // Required
 		TaskDefinition:          aws.String(taskDefName), // Required
 		Cluster:                 aws.String(client.params.Cluster),
 		DeploymentConfiguration: deploymentConfig,
-	})
-	if err != nil {
+	}
+
+	if *loadBalancer.TargetGroupArn != "" || *loadBalancer.LoadBalancerName != "" {
```
yinshiua: * I understand that for the current use case, the loadBalancer will always be set and will never be nil, but this `CreateServce` method can be reused in the future, hence I would still recommend adding a nil pointer check here.

* Can you change `*loadBalancer.TargetGroupArn` to use aws.StringValue (`aws.StringValue(loadBalancer.TargetGroupArn)` and `aws.StringValue(loadBalancer.LoadBalancerName)`)

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,35 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
 		},
 	}
 }
+
+func loadBalancerFlags() []cli.Flag {
+	targetGroupArnUsageString := "[Optional] The full Amazon Resource Name (ARN) of the Elastic Load Balancing target group associated with a service."
```
yinshiua: I am working on reviewing the text with others, and will update you once I have the final version.

ecs-cli/modules/compose/ecs/service.go
victorcoder: What I did in with the other params is let the AWS API validate the input of the user, the validation errors are quite meaninfull in most of the cases.

If we remove the `targetGoupArn`/`loadBalancerName` fallback here we get a good server side message educating the user: 

`FATA[0000] InvalidParameterException: loadBalancerName and targetGroupArn cannot both be specified. You must specify either a load BalancerName or a targetGroupArn.`

So removing the checking here

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,34 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
+	containerNameUsageString := "[Optional] Mandatory if target-group-arn is specified. The container name (as it appears in a container definition)."
+	containerPortUsageString := "[Optional] Mandatory if target-group-arn is specified. The container port to access from the load balancer."
+	loadBalancerNameUsageString := "[Optional] The name of the load balancer."
+	roleUsageString := fmt.Sprintf("[Optional] The name or full Amazon Resource Name (ARN) of the IAM role that allows Amazon ECS to make calls to your load balancer on your behalf. This parameter is required if you are using a load balancer with your service. If you specify the role parameter, you must also specify a target-group-arn, container-name and container-port.")
```
nrdlngr: Would you mind changing this to:

"[Optional] Specifies the name or full Amazon Resource Name (ARN) of the IAM role that allows Amazon ECS to make calls to your load balancer or target group on your behalf. This parameter is required if you are using a load balancer or target group with your service. If you specify the role parameter, you must also specify a load balancer name or target group ARN, along with a container name and container port."

ecs-cli/modules/compose/cli/ecs/app/service_command.go
nrdlngr: Would you mind changing this to:

"[Optional] Specifies the full Amazon Resource Name (ARN) of a previously configured Elastic Load Balancing target group to associate with your service."

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,34 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
 		},
 	}
 }
+
+func loadBalancerFlags() []cli.Flag {
+	targetGroupArnUsageString := "[Optional] The full Amazon Resource Name (ARN) of the Elastic Load Balancing target group associated with a service."
+	containerNameUsageString := "[Optional] Mandatory if target-group-arn is specified. The container name (as it appears in a container definition)."
```
nrdlngr: Would you mind changing this to:

"[Optional] Specifies the container name (as it appears in a container definition). This parameter is required if a load balancer or target group is specified."

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,34 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
+	containerPortUsageString := "[Optional] Mandatory if target-group-arn is specified. The container port to access from the load balancer."
```
nrdlngr: Would you mind changing this to:

"[Optional] Specifies the port on the container to associate with the load balancer. This port must correspond to a containerPort in the service's task definition. This parameter is required if a load balancer or target group is specified."

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -136,3 +136,34 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
+	loadBalancerNameUsageString := "[Optional] The name of the load balancer."
```
nrdlngr: Would you mind changing this to:

"[Optional] Specifies the name of a previously configured Elastic Load Balancing load balancer to associate with your service."

yinshiua:
@victorcoder Can you please make text changes base on @nrdlngr's suggestions? Thank you!

victorcoder:
Changes applied @nrdlngr

ecs-cli/modules/compose/cli/ecs/app/service_command.go
```diff
@@ -138,11 +138,11 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
+	containerNameUsageString := "[Optional] Specifies the container name (as it appears in a container definition). This parameter is required if a load balancer or target group is specified."
+	containerPortUsageString := "[Optional] Specifies the port on the container to associate with the load balancer. This port must correspond to a containerPort in the service's task definition. This parameter is required if a load balancer or target group is specified."
+	loadBalancerNameUsageString := "[Optional] Specifies the name of a previously configured Elastic Load Balancing load balancer to associate with your service."
+	roleUsageString := fmt.Sprintf("[Optional] Specifies the name or full Amazon Resource Name (ARN) of the IAM role that allows Amazon ECS to make calls to your load balancer or target group on your behalf. This parameter is required if you are using a load balancer or target group with your service. If you specify the role parameter, you must also specify a load balancer name or target group ARN, along with a container name and container port.")
```
yinshiua: Thanks for updating the text. Just one minor comment

printf is not longer needed here.

yinshiua:
@victorcoder  Are you fine with your commit being merged in under the Apache 2.0 license?
Thanks!

victorcoder:
@yinshiua Thanks to you, I'm ok, go ahead.

yinshiua:
Thanks for your contribution.

I am giving a pass on this PR, but will have minor followup changes on bug fixing.

markriggins:
Shouldn't this be able to read from a JSON or YAML file, so the service can be created, dropped, recreated without having to reproduce command line options?   

What about other service parameters such as maximumPercent and minimumHealthyPercent?

victorcoder:
@markriggins That could be a good improvement but those parameters wouldn't be compose compatible so it will force to have another file to create the service.

In any case, that would be the goal of another PR.

markriggins:
Yeah, we wound up adding a *service-name--env.json *file to our services,
 and then wrapping ecs-cli with a script to pick up these extra service
settings.    But we had no choice, since we could not add new keywords to
the yaml.    ecs-cli could add some custom sections, no?

Mark Riggins

CLOUD COMPUTING ARCHITECT

336.312.7125

www.SocialCode.com <http://socialcode.com/>

serragnoli:
Would it be possible for someone to give an example on how to use this new feature please?

yinshiua:
@markriggins This would be a good enhancement to read all these fields from yaml or JSON. (ex.`ecs-cli compose --f docker-compose.yml service up --f ecs-service-config.yml`)

@serragnoli An example for ALB would be: 
```
ecs-cli compose service up --target-group-arn arn:aws:elasticloadbalancing:us-west-2:<**aws-account-id**>:targetgroup/ecs-cli/090481523c5c99ea
--container-name wordpress --container-port 80 --role ecsServiceRole
```

markriggins:
@serragnoli <https://github.com/serragnoli>  -- this feature is currently
unavailable via ecs-cli, so we had to use the `aws ecs` command instead
after reading these settings from a json file and merging them with the
existing settings.

The important line from the script reads something like this.

    aws ecs update-service --service  ecscompose-service-$serviceName
--cli-input-json
"$service_settings"


Mark Riggins

CLOUD COMPUTING ARCHITECT

336.312.7125

www.SocialCode.com <http://socialcode.com/>

vchan2002:
@yinshiua Thank you for the example....

Just to clarify, when using the --target-group-arn setting, I would assume that the target group is already existing as a part of an ALB, and thus there's no need to specify the ALB, right?

Thus, when using the --load-balancer-name setting instead, it would assume that it is using a classic load balancer, not an ALB, since, as I'd imagine, just putting an ALB would be insufficient in stating where it needs to go?

I'll still trying to navigate my way thru in terms of understanding ECS, in terms of tasks, services, etc... so any help would be greatly appreciated!

yinshiua:
@vchan2002 
Yes, current feature only allows you to use an existing ALB or ELB.
To specify an ALB, you use `--target-group-arn`
To specify an ELB, you use `--load-balancer-name`
Once the service is created with an ALB or ELB, the name/arn is immutable to the service. You will not be able to switch to use a different ALB or ELB.

The following documentations may be helpful:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-load-balancing.html
http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_LoadBalancer.html

PR #220, opened 2017-03-15
Adding r4 instances to cloudformation template

antdrums:
The [r4](https://aws.amazon.com/about-aws/whats-new/2016/11/introducing-amazon-ec2-r4-instances-the-next-generation-of-memory-optimized-instances/) instance type doesn't seem to be supported by the cli / sdk. Updating the template file to support the following : 
r4.large
r4.xlarge
r4.2xlarge
r4.4xlarge
r4.8xlarge
r4.16xlarge


sagaramagi:
has this been merged into master?

Issue #57, opened 2016-01-06
compose different environments


esetnik:
`docker-compose` expose the `-f` switch to load one or more docker-compose override files. Is there something similar for ecs-cli?

floostmodern:
It seems to me that #129 is a duplicate of this. It would be really great to make `ecs-cli` support multiple `-f`.

In the meantime, there is a workaround : `docker-compose` has a validation command that shows the yaml files' concatenation in the console :

`docker-compose -f docker-compose.file1.yml -f docker-compose.file2.yml config` will output the concatenation of ` docker-compose.file1.yml` and `docker-compose.file2.yml`.

chrisckc:
Hi, I have just started using AWS for running docker and this is the first issue i encountered, is there any update on this?
The workaround suggested by @floostmodern is good and can be put in a bash script to capture the output to a new file and execute the  ecs-cli command with that file, but it would be better if the workaround was not required.
Thanks

PeterRJones:
While it is true that `docker-compose -f docker-compose.file1.yml -f docker-compose.file2.yml config` will output the merged config for all the given override files, the resulting file: 
- is not valid YAML
- is not valid docker-compose format
- is not consumable by ecs-cli
- requires additional touch-ups before it can be successfully to `ecs-cli`

I don't really see this as a valid work-around for this. If this feature is not to be supported, it would be *helpful* to show an error when ecs-cli is run with this type of syntax (today this is allowed, seems to work but silently fails):

jdno:
We require this functionality as well to set a few environment/stage/cluster specific options for our containers. While we could potentially configure them with environment variables as well, it is much more inline with our development workflow to merge and overwrite them using different files.

We normally run services like this:
    $ docker-compose -f service.yml -f stage.yml up/run

While the workaround by @floostmodern works for now, it would be great if `ecs-cli` supported this functionality natively. Who knows if `docker-compose config` will work forever...

AlJohri:
Building on what @floostmodern said, this works for me temporarily.

Full Command:
```bash
docker-compose -f docker-compose.yml -f docker-compose.production.yml config | sed "s/version: '2.0'/version: 2/" | ecs-cli compose --verbose --file /dev/stdin create
```

I tend to shorten it like so:
```bash
alias docker-compose-prod='docker-compose -f docker-compose.yml -f docker-compose.production.yml'
[...]

PR #229, opened 2017-04-04
Proposal: compose run supports docker-compose style arguments

akirakoyasu:
Addresses Issue #148
reference #150

`compose run` supports multi-argument commands as:
`ecs-cli compose run <container> <cmd0> <cmd1> <cmd2>`

It's a breaking change.

### DIFF
--- a/ecs-cli/modules/aws/clients/ecs/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/client.go
@@ -61,7 +61,7 @@ type ECSClient interface {
 	// Tasks related
 	GetTasksPages(listTasksInput *ecs.ListTasksInput, fn ProcessTasksAction) error
 	RunTask(taskDefinition, startedBy string, count int) (*ecs.RunTaskOutput, error)
-	RunTaskWithOverrides(taskDefinition, startedBy string, count int, overrides map[string]string) (*ecs.RunTaskOutput, error)
+	RunTaskWithOverrides(taskDefinition, startedBy string, count int, container string, commands []string) (*ecs.RunTaskOutput, error)
 	StopTask(taskId string) error
 	DescribeTasks(taskIds []*string) ([]*ecs.Task, error)
 
@@ -398,16 +398,13 @@ func (client *ecsClient) RunTask(taskDefinition, startedBy string, count int) (*
 }
 
 // RunTask issues a run task request for the input task definition
-func (client *ecsClient) RunTaskWithOverrides(taskDefinition, startedBy string, count int, overrides map[string]string) (*ecs.RunTaskOutput, error) {
+func (client *ecsClient) RunTaskWithOverrides(taskDefinition, startedBy string, count int, container string, commands []string) (*ecs.RunTaskOutput, error) {
 
-	commandOverrides := []*ecs.ContainerOverride{}
-	for cont, command := range overrides {
-		contOverride := &ecs.ContainerOverride{
-			Name:    aws.String(cont),
-			Command: aws.StringSlice([]string{command}),
-		}
-		commandOverrides = append(commandOverrides, contOverride)
-	}
+	commandOverrides := []*ecs.ContainerOverride{
+		{
+			Name:    aws.String(container),
+			Command: aws.StringSlice(commands),
+		}}
 	ecsOverrides := &ecs.TaskOverride{
 		ContainerOverrides: commandOverrides,
 	}
--- a/ecs-cli/modules/aws/clients/ecs/mock/client.go
+++ b/ecs-cli/modules/aws/clients/ecs/mock/client.go
@@ -193,15 +193,15 @@ func (_mr *_MockECSClientRecorder) RunTask(arg0, arg1, arg2 interface{}) *gomock
 	return _mr.mock.ctrl.RecordCall(_mr.mock, "RunTask", arg0, arg1, arg2)
 }
 
-func (_m *MockECSClient) RunTaskWithOverrides(_param0 string, _param1 string, _param2 int, _param3 map[string]string) (*ecs.RunTaskOutput, error) {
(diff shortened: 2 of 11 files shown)


yinshiua:
ECS works slightly different from docker. docker-compose can run `compose run` at the container level, so they can configure one container at a time, where as ECS has to run `compose run` on multiple containers. This is not just a backward incompatible change, this change is also taking away the ability to configure commands on multiple containers.
I'll come up with a proposal by next week.

yinshiua:
Here are some ideas. Let me know what you think.

## Option 1
```
ecs-cli compose run <container> <cmd0> <cmd1> <cmd2>...
ecs-cli compose run <container> --override override.yaml
```
Pros:
* consistent with docker-compose run command

Cons:
* If you want to have multiple commands, you will need to use a yaml file
* backward incompatible

## Option 2
```
ecs-cli compose run <container1> “<cmd0> <cmd1> <cmd2>” <container2> “<cmd0> <cmd1> <cmd2>...”
```
Pros:
* Backward compatible
* Easy to use, does not need to create extra files

Cons:
* inconsistent with docker-compose
* The command can get very long

## Option 3

```
ecs-cli compose run <container1>=“<cmd0> <cmd1> <cmd2>” <container2>=“<cmd0> <cmd1> <cmd2>...”
```
Pros:
* Easier to read than option2

Cons:
* Still not backward compatible

akirakoyasu:
@yinshiua Thanks!
I consider one of the most important `ecs-cli compose`'s benefits is interoperability with docker-compose. Most of the `ecs-cli compose` users probably expect the behavior like docker-compose. The docker-compose way to override service definition is multiple `--file`.

I propose Option 1' that supports multiple `--file` and docker-compose compatible `run`.

## Option 1'
```
ecs-cli compose --file <yaml1> --file <yaml2>... run <container> <cmd0> <cmd1> <cmd2>...
```

Pros:
- consistent with docker-compose run command

Cons:
- If we want to have multiple commands, we will need to use a yaml file (docker-compose yaml)
- backward incompatible

----

And I suggest `ecs-cli compose` commands' design refer to docker-compose as much as possible. Commands and options which docker-compose doesn't have are possible inconsistency with an update of docker-compose in the future.

yinshiua:
**docker-compose override** will actually create a task definition which will set the default command to whatever's in the override yaml (one use case is if you have different deployment stages: https://github.com/aws/amazon-ecs-cli/issues/57)
**run command override** is only used once at runtime and will not create a new task definition. You will be able to reuse the same task definition with different command overrides.

The two are slightly different use cases, **docker compose override** is more permanent whereas **run command override** is more for one time use.
***
I would break this problem down into two parts:
* docker-compose override 
```
ecs-cli compose --file <yaml1> --file <yaml2> run
```

* run command override (we would still need a way to set run command overrides on multiple containers)
```
ecs-cli compose run <container1> “<cmd0> <cmd1> <cmd2>” <container2> “<cmd0> <cmd1> <cmd2>...”
```


***
Eventually combining both would be
```
ecs-cli compose -f <yaml> -f <yaml> run <container1> “<cmd0> <cmd1> <cmd2>” <container2> “<cmd0> <cmd1> <cmd2>...”
```
but, depending on your use case, we should try to tackle one problem first. From my understanding **run command override** is generally used when you're trying to debug a container.

Can you elaborate more on what your use case is? Are you trying to run commands for different development stage? or Are you trying to debug your container? or Do you have other use cases

akirakoyasu:
I'm trying to run several jobs from a workflow system.

A few types of our runtime environments are containerized, and the workflow system kicks jobs in the container with some parameters. The parameters specify the job input / output or configuration files, so that have many combinations.

I plan to define the containerized runtime environments as task definitions and to pass the parameters to 'run command override'.

yinshiua:
Got it. We should then focus on the **run command override** solution.

> ```
> ecs-cli compose run <container1> “<cmd0> <cmd1> <cmd2>” <container2> “<cmd0> <cmd1> <cmd2>...”
> ```

You will need to use github.com/flynn/go-shlex to parse the command arguments. amazon-ecs-cli already has dependency on this package.

akirakoyasu:
ok. I'll implement it and push by next week.

pgeez:
@yinshiua I'm late to the game, but why are we breaking compatibility with Docker Compose by opting for Option 2? Isn't Option 1 the most consistent with the existing Docker Compose `run` syntax?

https://docs.docker.com/compose/reference/run/

@akirakoyasu Same question. You mentioned that interoperability between ECS-CLI and Docker Compose is very important. If so, why wouldn't you just implement Option 1? Doesn't Docker Compose `run` only operate on one service/container at a time?

yinshiua:
@pgeez As I mentioned above,  Docker runs at the container level whereas ECS runs at the task level.
> ECS works slightly different from docker. docker-compose can run `compose run` at the container level, so they can configure one container at a time, where as ECS has to run `compose run` on multiple containers.

Option1 will take away the ability to configure other containers if your ECS task is running multiple containers, so it's not doable in the ECS world.

akirakoyasu:
@yinshiua I implemented it in #235.

@pgeez I agree with you on the interoperability importance. But I'd like to deal with this `compose run` capability of multi-argument now.

@yinshiua @pgeez
ECS-CLI vs Docker Compose mismatch is a big problem. Why don't we discuss it in the issue #148 ?

yinshiua:
@akirakoyasu thanks for the new PR.

Because of the limitations in ECS, we don't support what Docker Compose does today. You can still reuse the commands that you used with docker-compose you just need to add quotes around the commands in order to use in the ecs-cli

pgeez:
Does closing this PR mean that ECS-CLI is not going to support multiple run
commands?

yinshiua:
New PR created https://github.com/aws/amazon-ecs-cli/pull/235

Issue #231, opened 2017-04-11
ecs-cli ps [signal SIGSEGV: segmentation violation code=0x1 addr=0xe8 pc=0x9f5918]


ricardson:
When I run **ecs-cli ps** in one of my ECS Clusters I get the error, to the others ECS Clusters works just fine.

[signal SIGSEGV: segmentation violation code=0x1 addr=0xe8 pc=0x9f5918]

goroutine 1 [running]:
github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs.getContainersForTasks(0x11e78e0, 0xc4201ac930, 0xc4201abf40, 0x6, 0x8, 0x8, 0x0, 0x0, 0x0, 0xc420260b20)
	/usr/src/app/src/github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs/entity.go:195 +0x498
github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs.collectContainers(0x11e78e0, 0xc4201ac930, 0xc4201e8100, 0x0, 0x2, 0xc4204979d0, 0x40f348, 0x0)
	/usr/src/app/src/github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs/entity.go:113 +0xc2
github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs.info(0x11e78e0, 0xc4201ac930, 0x9f9800, 0xbc7c40, 0x1232e18, 0xc4201ac930, 0xc420260b20, 0x11dffa0)
	/usr/src/app/src/github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs/entity.go:98 +0x42
github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs.(*Task).Info(0xc4201ac930, 0x11e7800, 0xc4201ac930, 0xc42019fb60, 0x0, 0x0, 0x40f348)
	/usr/src/app/src/github.com/aws/amazon-ecs-cli/ecs-cli/modules/compose/ecs/task.go:101 +0x40
github.com/aws/amazon-ecs-cli/ecs-cli/modules/command.clusterPS(0xc4200d3b80, 0x11e44c0, 0xc4201b6c00, 0x11e7fa0, 0xc42019fb60, 0xcd088d, 0x5, 0x4, 0xc4201b2be8, 0x5)
[...]

PR #234, opened 2017-04-13
Fix segmentation violation error when ec2InstanceId does not exist

yinshiua:
closes aws/amazon-ecs-cli#231.

`ecs-cli ps` gives segmentation violation error because it is trying to get the `PublicIPAddress` using the `EC2InstanceID` provided by the container instance, but the EC2 instance no longer exist.

This can be reproduced by first **stop** then **terminate** the EC2 instance. This error should go away without the fix ECS container instance recognize that the EC2 instance has been terminated.

### DIFF
--- a/ecs-cli/modules/compose/ecs/entity.go
+++ b/ecs-cli/modules/compose/ecs/entity.go
@@ -183,7 +183,7 @@ func getContainersForTasks(entity ProjectEntity, ecsTasks []*ecs.Task) ([]Contai
 	}
 
 	ec2InstanceIds := listEC2Ids(containerToEC2InstanceIDs)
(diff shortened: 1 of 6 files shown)


ecs-cli/modules/compose/ecs/entity.go
```diff
@@ -183,15 +183,17 @@ func getContainersForTasks(entity ProjectEntity, ecsTasks []*ecs.Task) ([]Contai
 	}
 
 	ec2InstanceIds := listEC2Ids(containerToEC2InstanceIDs)
+	fmt.Println(ec2InstanceIds)
```
uttarasridhar: lets remove print statements

PR #243, opened 2017-05-03
Use assert in cluster, config, and params tests

yinshiua:
This is the first commit of PR https://github.com/aws/amazon-ecs-cli/pull/242

### DIFF
--- a/ecs-cli/modules/config/destination_test.go
+++ b/ecs-cli/modules/config/destination_test.go
@@ -16,6 +16,7 @@ package config
 import (
 	"io/ioutil"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


ecs-cli/modules/config/destination_test.go
```diff
@@ -32,15 +33,7 @@ func TestNewDefaultDestination(t *testing.T) {
...
-
-	if !strings.HasSuffix(dest.Path, "ecs") {
-		t.Errorf("Invalid suffix for ecs config path: ", dest.Path)
-	}
-
-	if !dest.Mode.IsDir() {
-		t.Errorf("user home directory expected to have directory in mode")
-	}
+	assert.NoError(t, err, "Unexpected error creating new config path")
+	assert.Contains(t, dest.Path, "ecs", "Invalid suffix for ecs config path")
```
uttarasridhar: Instead of checking for just contains, can we verify hasSuffix

PR #245, opened 2017-05-10
modify the template to support additional instance types p2, g2, x1

zachahuy:
These instances are currently not supported through ecs-cli for up. It would be useful for people running GPU or large database workloads.


joereyna:
+1

Issue #188, opened 2017-01-04
How to specify taskRoleArn in the .yml file?


rclmenezes:
The ecs TaskDefinition API allows you to specify the exact IAM role for your task:
http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskDefinition.html

However, the docs don't specify if there's a way to do this in ecs-cli.

kux:
+1 for this feature request

kux:
In the meanwhile, a workaround is the fact that the containers inherit the policy that the role of the ec2 host has.

ogrodnek:
@kux ec2 instance roles are better than nothing, but it's not really a workaround.  with ec2 roles all tasks running on that instance will have the host's role vs. task specific policies.

BlizzardBlue:
+1

mihbor:
+1

vbudilov:
+1
@yinshiua @uttarasridhar 

Any idea if this is on the roadmap? It's an important security feature. 

There's also no way to specify the role on the EC2 instances themselves with *ecs-cli up* (right now I have to assign different roles or add policies after the instances have been provisioned), so it's a double whammy.

yinshiua:
Hi, these are the features that we definitely want to support. We're still working through our priorities, and will keep you posted. Feel free to submit a PR.

vbudilov:
Hi Jessica, a PR was submitted for this - I linked to our in my previous
comment.

yinshiua:
I've commented on the PR :)

Feel free to submit a PR for instance role as well. Thanks!

yinshiua:
You can now specify task role arn at runtime using `--task-role-arn`, but we will need a different solution as we add more ECS specific Task definition fields that are not supported by Docker compose in either the docker-compose yml or a separate ECS yaml.

adenot:
> You can now specify task role arn at runtime using --task-role-arn, but we will need a different solution as we add more ECS specific Task definition fields that are not supported by Docker compose in either the docker-compose yml or a separate ECS yaml.

We could use labels in the compose yaml file to achieve this:

```
  labels:
    aws.ecs.task.role_arn: 'some_arn_here'
```

jlyon:
@yinshiua `--task-role-arn` is not working for me in `:latest` (`VERSION:  0.6.0 (1053c82)`):
```
Incorrect Usage: flag provided but not defined: -task-role-arn
```
[...]

PR #247, opened 2017-05-12
Adds support for --task-role-arn with compose commands

rifelpet:
This addresses #188 by providing a `--task-role-arn` flag with the compose command. The TaskRoleArn is passed to the RegisterTaskDefinition API call. Storing it as an property in the yml file is not an option since it would require modifying libCompose, so I opted for the command line argument instead.

Let me know if you'd like me to make any adjustments before merging.

### DIFF
--- a/ecs-cli/modules/cli/compose/project/project.go
+++ b/ecs-cli/modules/cli/compose/project/project.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/amazon-ecs-cli/ecs-cli/modules/utils/compose"
 	"github.com/docker/libcompose/config"
 	"github.com/docker/libcompose/project"
+	"github.com/aws/amazon-ecs-cli/ecs-cli/modules/commands"
 )
 
 // Project is the starting point for the compose app to interact with and issue commands
@@ -134,7 +135,8 @@ func (p *ecsProject) transformTaskDefinition() error {
 	// convert to task definition
 	logrus.Debug("Transforming yaml to task definition...")
 	taskDefinitionName := utils.GetTaskDefinitionName(context.ECSParams.ComposeProjectNamePrefix, context.Context.ProjectName)
-	taskDefinition, err := utils.ConvertToTaskDefinition(taskDefinitionName, &context.Context, p.ServiceConfigs())
+	taskRoleArn := context.CLIContext.GlobalString(command.TaskRoleArnFlag)
+	taskDefinition, err := utils.ConvertToTaskDefinition(taskDefinitionName, &context.Context, p.ServiceConfigs(), taskRoleArn)
 	if err != nil {
 		return err
 	}
--- a/ecs-cli/modules/commands/compose/compose_command.go
+++ b/ecs-cli/modules/commands/compose/compose_command.go
@@ -90,6 +90,10 @@ func composeFlags() []cli.Flag {
 			Usage:  "Specifies the project name to use. Defaults to the current directory name.",
 			EnvVar: "COMPOSE_PROJECT_NAME",
 		},
+		cli.StringFlag{
+			Name:   command.TaskRoleArnFlag,
+			Usage:  "[Optional] Specifies the short name or full Amazon Resource Name (ARN) of the IAM role that containers in this task can assume. All containers in this task are granted the permissions that are specified in this role.",
+		},
 	}
 }
 
--- a/ecs-cli/modules/cli/compose/entity/entity_helper.go
+++ b/ecs-cli/modules/cli/compose/entity/entity_helper.go
@@ -56,6 +56,7 @@ func GetOrCreateTaskDefinition(entity ProjectEntity) (*ecs.TaskDefinition, error
 		Family:               taskDefinition.Family,
 		ContainerDefinitions: taskDefinition.ContainerDefinitions,
 		Volumes:              taskDefinition.Volumes,
+		TaskRoleArn:          taskDefinition.TaskRoleArn,
 	}, entity.TaskDefinitionCache())
 
 	if err != nil {
--- a/ecs-cli/modules/commands/flags.go
+++ b/ecs-cli/modules/commands/flags.go
@@ -55,6 +55,7 @@ const (
 	// Compose
 	ProjectNameFlag     = "project-name"
 	ComposeFileNameFlag = "file"
+	TaskRoleArnFlag     = "task-role-arn"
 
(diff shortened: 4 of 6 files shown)


ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -102,6 +104,7 @@ func (s *Service) LoadContext() error {
 		}
 		s.role = role
 	}
+	s.taskRoleArn = s.Context().TaskRoleArn
```
yinshiua: s.taskRoleArn = s.Context().CLIContext.String(command.TaskRoleArnFlag)

ecs-cli/modules/cli/compose/context/context.go
```diff
@@ -40,7 +40,8 @@ type Context struct {
 	EC2Client ec2client.EC2Client
 
 	// IsService would decide if the resource created by this compose project would be ECS Tasks directly or through ECS Services
-	IsService bool
+	IsService   bool
+	TaskRoleArn string
```
yinshiua: Not the best place to put this.

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -41,6 +41,7 @@ type Service struct {
 	deploymentConfig *ecs.DeploymentConfiguration
 	loadBalancer     *ecs.LoadBalancer
 	role             string
+	taskRoleArn      string
```
yinshiua: This should be part of taskDef. ecs.TaskDefinition has a TaskRoleArn field

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -55,6 +56,7 @@ func NewService(context *context.Context) entity.ProjectEntity {
 		cache:          entity.SetupTaskDefinitionCache(),
 		projectContext: context,
 		timeSleeper:    &utils.TimeSleeper{},
+		taskRoleArn:    context.TaskRoleArn,
```
yinshiua: TaskRoleArn is used when creating task definition, not used for create service hence doesn't make sense to be part of the service object.

ecs-cli/modules/cli/compose/context/context.go
hasPeterr: @yinshiua I agree, however I wasn't sure where to put it since it can't go in `libcompose`'s `Project`. Do you have any suggestions on where it should go?

yinshiua: Good question! Eventually, we will need either a wrapper around the libcompose context or another ECS task definition context to support ECS task definition fields that are not supported by libcompose.

For now, we can directly add `taskRokeArn := context.CLIContext.String(command.TaskRoleArnFlag)` in project.go transformTaskDefinition function.

rifelpet:
@yinshiua I've addressed your comments and squashed the commits.

joaonlforte:
This would be extremely helpful!! Are there any plans to release this shortly? I would like to avoid all the hassle of having to use aws cli with a baked .json to be able to set the task role :s using the compose would be a much nicer solution

ecs-cli/modules/commands/compose/compose_command.go
```diff
@@ -90,6 +90,11 @@ func composeFlags() []cli.Flag {
 			Usage:  "Specifies the project name to use. Defaults to the current directory name.",
 			EnvVar: "COMPOSE_PROJECT_NAME",
 		},
+		cli.StringFlag{
+			Name:   command.TaskRoleArnFlag + ",r",
```
yinshiua: I don't think we want to provide the short handed flag here.

rifelpet:
Updated to not provide a short handed flag for `--task-role-arn`

ecs-cli/modules/cli/compose/project/project.go
```diff
@@ -134,7 +135,8 @@ func (p *ecsProject) transformTaskDefinition() error {
 	// convert to task definition
 	logrus.Debug("Transforming yaml to task definition...")
 	taskDefinitionName := utils.GetTaskDefinitionName(context.ECSParams.ComposeProjectNamePrefix, context.Context.ProjectName)
-	taskDefinition, err := utils.ConvertToTaskDefinition(taskDefinitionName, &context.Context, p.ServiceConfigs())
+	taskRoleArn := context.CLIContext.String(command.TaskRoleArnFlag)
```
yinshiua: This needs to be `context.CLIContext.GlobalString` since the command is going to be `ecs-cli compose --task-role-arn <arn> service up`

ecs-cli/modules/commands/compose/compose_command.go
```diff
@@ -90,6 +90,11 @@ func composeFlags() []cli.Flag {
 			Usage:  "Specifies the project name to use. Defaults to the current directory name.",
 			EnvVar: "COMPOSE_PROJECT_NAME",
 		},
+		cli.StringFlag{
+			Name:   command.TaskRoleArnFlag,
+			Usage:  "Specifies the TaskRoleArn to use with the task definition.",
+			EnvVar: "TASK_ROLE_ARN",
```
yinshiua: We don't really have implementation on reading from EnvVar here. I would remove this for now and we can come back to this when we have a full solution on how we want to handle ECS TaskDef fields that are not supported by compose.

hasPeterr:
Removed the environment variable usage and switched to `CLIContext.GlobalString`

rifelpet:
Removed environment variable usage and switched to `CLIContext.GlobalString`

ecs-cli/modules/commands/compose/compose_command.go
```diff
@@ -90,6 +90,10 @@ func composeFlags() []cli.Flag {
 			Usage:  "Specifies the project name to use. Defaults to the current directory name.",
 			EnvVar: "COMPOSE_PROJECT_NAME",
 		},
+		cli.StringFlag{
+			Name:   command.TaskRoleArnFlag,
+			Usage:  "Specifies the TaskRoleArn to use with the task definition.",
```
nrdlngr: [Optional] Specifies the short name or full Amazon Resource Name (ARN) of the IAM role that containers in this task can assume. All containers in this task are granted the permissions that are specified in this role.

rifelpet:
Documentation text updated.

rifelpet:
Yes I am!

Issue #176, opened 2016-11-02
`ecs-cli configure --cluster C`  erases all the other settings


markriggins:
Is this intentional?    Should `ecs-cli configure --cluster C` erase all the other configuration values?

**How to reproduce this problem:**
create a default config file with a cluster and region
```
$ echo '' >home/.ecs/config
$ HOME=$PWD/home ecs-cli.0.4.4 configure --cluster a-really-long-long-name --region us-east-1
$ cat home/.ecs/config
[ecs]
cluster                     = a-really-long-long-name
aws_profile                 = 
region                      = us-east-1
aws_access_key_id           = 
aws_secret_access_key       = 
compose-project-name-prefix = ecscompose-
compose-service-name-prefix = ecscompose-service-
cfn-stack-name-prefix       = amazon-ecs-cli-setup-
```

Now try to change the cluster:
```
$ HOME=$PWD/home ecs-cli.0.4.4 configure --cluster a-really-longer-longer-longer-longer-name
$ cat home/.ecs/config
[ecs]
cluster                     = a-really-longer-longer-longer-longer-name
aws_profile                 = 
region                      = 
aws_access_key_id           = 
aws_secret_access_key       = 
compose-project-name-prefix = ecscompose-
compose-service-name-prefix = ecscompose-service-
cfn-stack-name-prefix       = amazon-ecs-cli-setup-
```

And we see that it has forgotten the **region**.   

Do we really have to give you **EVERY ARGUMENT** every time we use the `ecs-cli configure` command?

yinshiua:
Added proposals to support multiple clusters configuration. Please give us some feedback. Thanks! https://github.com/aws/amazon-ecs-cli/issues/253

yinshiua:
Released in v1.0.0

Issue #152, opened 2016-08-30
Swapping between clusters nukes credentials


kaihendry:
ecs-cli version 0.4.3 (c6ea81e)

```
~/.ecs$ cat config
[ecs]
cluster                     = website
region                      = ap-southeast-1
aws_access_key_id           = MYID
aws_secret_access_key       = SECRET
compose-project-name-prefix = ecscompose-
compose-service-name-prefix = ecscompose-service-
cfn-stack-name-prefix       = amazon-ecs-cli-setup-
aws_profile                 = spuul
ame-prefix                  = amazon-ecs-cli-setup-

~/.ecs$ ecs-cli configure -c tv-app
INFO[0000] Saved ECS CLI configuration for cluster (tv-app)
~/.ecs$ cat config
[ecs]
cluster                     = tv-app
region                      =
aws_access_key_id           =
aws_secret_access_key       =
compose-project-name-prefix = ecscompose-
compose-service-name-prefix = ecscompose-service-
cfn-stack-name-prefix       = amazon-ecs-cli-setup-
aws_profile                 =
ame-prefix                  = amazon-ecs-cli-setup-

le                 = spuul
ame-prefix                  = amazon-ecs-cli-setup-
```

I think the workaround confusingly is `COMPOSE_PROJECT_NAME`, but I'm not 100% sure.

pgeez:
I think this is related to #59 

👍

joeboyscout04:
And the switching clusters also puts some junk at the bottom of the file, which causes "Error initializing: key-value delimiter not found"

yinshiua:
related to https://github.com/aws/amazon-ecs-cli/issues/174

yinshiua:
Added proposals to support multiple clusters configuration. Please give us some feedback. Thanks! https://github.com/aws/amazon-ecs-cli/issues/253

yinshiua:
Released in v1.0.0

Issue #112, opened 2016-05-19
Configure profiles per directory and/or extract from ~/.aws/config file


Dudemullet:
I thinks it's kind of bothersome to have to run `ecs-cli configure -p [profile] --cluster [cluster-name]` everytime I ump between project directories. Not only that but, why not save the `cluster` in the `~/.aws/config` profile, and just extract it from there when using `ecs-cli configure -p [profile]` ?

I don't know if I'm missing some sort of configuration shortcut or whatnot, but if anyone else has this issues and actually running `ecs-cli configure` is the only solution we have right now, I'd happily submit a PR for something like:
## Proposal 1

`ecs-cli configure -p [profile]` - Extracts all the needed variables in the `~/.aws/config` [profile] to set in `~/.ecs/config` like:
- cluster
- aws_profile
- region
- aws_access_key_id
- aws_secret_access_key
- compose-project-name-prefix
- compose-service-name-prefix
- cfn-stack-name-prefix
## Proposal 2

running `ecs-cli configure` on a directory will try to populate the configuration via a projects local `./.ecs/config` file.

MaerF0x0:
# Proposal 3

`ecs-cli <command> --cluster [name]`

skwp:
Huge 👍 having the cluster as a global config makes it impossible to deploy to multiple ecs clusters in a pipeline environment like jenkins because each job would overwrite the global config as it deploys.

skwp:
Proposal 3 is the most sane to me. Each command should take --cluster as an arg and override the global.

sveneh:
I'm also in favor of proposal 3, but it also needs a global `--region` parameter for full flexibility

pjoe:
I vote for proposal 2 :)

It very much matches how eb-cli deals with this.

fdeschenes:
I also vote for proposal 2.

Dudemullet:
Hey all, I haven't given up on this issue. But after numerous attempts at trying to downgrade my go version (and not being able to get gvm running), understanding how Go's (and godep) dependency management work and most importantly trying to learn Go. It's proven pretty difficult for me to just jump in on this project and contribute. I will try on my christmas break to get back on board though.

But right now Go/godep has me like:

![jackie-chan-wtf](https://cloud.githubusercontent.com/assets/491242/21086057/80891190-bfde-11e6-8d7a-be5e864cb35e.jpg)

kidanekal:
I also vote for proposal 2.

[...]

Issue #59, opened 2016-01-07
Support multiple clusters in configuration


raphink:
The configuration file currently uses only one section (`ecs`).

It would be very useful if several sections could be used, one per cluster, with the ability to switch between them, e.g.:

``` shell
$ cat ~/.ecs/config
[ecs]
cluster = default-cluster
aws_profile = 
region = eu-west-1
aws_access_key_id = MYKEYID
aws_secret_access_key = mySecretKey

[test]
cluster = another-cluster
aws_profile =
region = us-east-1
aws_access_key_id = MYKEYID
aws_secret_access_key = mySecretKey

$ ecs-cli -c test ps
```

christianblunden:
Even better would be to be able to pass the cluster name at run time like the other attributes. eg:

aws-cli --cluster another-cluster --region eu-west-1 compose create

joprice:
👍

DylanBruzenak:
👍

pgeez:
👍

AlJohri:
👍

yinshiua:
related to https://github.com/aws/amazon-ecs-cli/issues/174

LavieX:
👍

yinshiua:
Added proposals to support multiple clusters configuration. Please give us some feedback. Thanks! https://github.com/aws/amazon-ecs-cli/issues/253

gstaples:
Is there any traction on this?  Why is this config file even a thing?  Just use --profile like every other AWS command and utility?

I deal with multiple accounts. I will have to deal with multiple clusters in multiple accounts.  Without this support, I'm not sure how to move forward.

yinshiua:
@gstaples We posted a proposal earlier #251, and we are actively working on the implementation.

As for today, added `--cluster` and `--region` runtime flags for every command in our latest release. You can use AWS CLI profile and the new runtime flags to switch between accounts and clusters. Hope that helps!

yinshiua:
Released in v1.0.0

PR #251, opened 2017-06-07
Add proposal for Multiple clusters configuration

yinshiua:
Adds a proposal document for multiple clusters configuration.

Related to #59, #71, #112, #152, #174, #176

Related to proposal https://github.com/aws/amazon-ecs-cli/pull/252

### DIFF
--- a/proposals/multi-cluster-config.md
+++ b/proposals/multi-cluster-config.md
@@ -1,0 +1,155 @@
+<!--
+Copyright 2015-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"). You may
+not use this file except in compliance with the License. A copy of the
+License is located at
+
+http://aws.amazon.com/apache2.0/
+
+or in the "license" file accompanying this file. This file is distributed
+on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+express or implied. See the License for the specific language governing
+permissions and limitations under the License.
+-->
+
+# Introduction
+ECS CLI currently only support one cluster configuration which results in a number of issues for the users:
+1. When customer switch clusters or regions within the same account, they need to reconfigure the setup.
+1. When customer is running two Jenkins builds simultaneously, the config file gets reconfigured while one job is still processing.
+1. When customers work as a team, each person has to run `ecs-cli configure` to configure the correct cluster and region.
+1. Everything (cluster name, region, creds) should be configurable via arguments or environment variables.
+
+# Solution
+When users run `ecs-cli configure`, ECS CLI will configure their cluster configuration in `~/.ecs/ecs.config`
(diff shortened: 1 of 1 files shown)


proposals/multi-cluster-config.md
```diff
@@ -0,0 +1,155 @@
...
+Does not need to configure cluster
+1. Configure ECS profile
+   ```
+   ecs-cli configure  profile --access-key ********** --secret-key ********** --profile-name my_profile
+   ```
+1. Cluster PS
+   ```
+   ecs-cli ps --cluster cli-demo --region us-west-2 --ecs-profile my_profile
+   ```
+1. Do the same with AWS profile
```
markriggins: 👍 -- that's what I really need for use on Jenkins, when we really don't want to contend with multiple projects that need to work with various clusters simultaneously

proposals/multi-cluster-config.md
```diff
@@ -0,0 +1,155 @@
...
+# Workflow
+### Single cluster first run
+1. Configure ECS credentials (Stores in `~/.ecs/ecs.profile`)
+   ```
+   ecs-cli configure  profile --access-key ********** --secret-key ********** --profile-name my_profile
+   ```
+1. Configure cluster (Stores in `~/.ecs/ecs.config`)
+   ```
+   ecs-cli configure --cluster cli-demo --region us-west-2 --config-name beta_config
+   ```
```
markriggins: it's not clear to me how the **config-name** of `beta_config` factors into this.   It does not appear to be a field in the ~/.ecs/ecs.config YAML.   Does this result in a  new`~/.ecs/beta_config` file, or a section in the ~/.ecs/ecs.config file?

yinshiua: We allow multiple cluster configurations in ~/.ecs/ecs.config, `--config-name` is a way to identify the different cluster profile which can be specified at runtime. Assume you have an existing cluster config "ecs" and you would like to configure a different cluster but not overwriting the previous one, you can run the command with `--config-name` with "beta_config" to add the new cluster in the cluster config.
```yaml
default: ecs
clusters:
  ecs: 
      cluster: ecs-test
      region: us-west-1
  beta_config:
      cluster: cli-demo
      region: us-west-2
```

markriggins: if these are "configurations" why are we listing them under "clusters:" in the above YAML?

how about simply doing this instead?
```
default: ecs-test
clusters:
   ecs:
      region: use-west-1
   cli-demo:
      region: us-west-2
```

proposals/multi-cluster-config.md
gstaples: Isn't it redundant to specify --region when it's already in the profile?  Why not just use --profile like every other AWS command?

yinshiua: --region is a way for you to overwrite your cluster configuration which comes in handy when you are running script to perform the same actions in multiple regions.

PR #252, opened 2017-06-07
Add proposal for removing prefixes from config

yinshiua:
Adds a proposal document for removing prefixes from config.

Related to proposal https://github.com/aws/amazon-ecs-cli/pull/251

### DIFF
--- a/proposals/remove-prefix-from-config.md
+++ b/proposals/remove-prefix-from-config.md
@@ -51,17 +51,24 @@ Assume project name is "myApp"
 1. Config help
    ```
    $ ecs-cli configure --help
+   
+   COMMANDS:
+   migrate    Moves your old configuration to new configuration
+   
+   OPTIONS:
    --cluster value, -c value            Specifies the ECS cluster name to use. If the cluster does not exist, it is created when you try to add resources to it with the ecs-cli up command. [$ECS_CLUSTER]
    --region value, -r value             Specifies the AWS region to use. If the AWS_REGION environment variable is set when ecs-cli configure is run, then the AWS region is set to the value of that environment variable. [$AWS_REGION]
    --cfn-stack-name value               [Optional] Specifies the name of AWS CloudFormation stack created on ecs-cli up. (default: "amazon-ecs-cli-setup-<cluster-name>")
-   --migrate                            [Optional] Moves your old configuration to new configuration
-   --force-migrate                      [Optional] Force move your old configuration to new configuration
    --compose-service-name-prefix value  [Deprecated] Specifies the prefix added to an ECS service created from a compose file. Format <prefix><project-name>. (default to empty)
+   
+   $ ecs-cli configure migrate --help
+   --force                      [Optional] Force move your old configuration to new configuration
+
    ```
 
 1. Migrate config with prompts
    ```
-   $ ecs-cli configure --migrate
+   $ ecs-cli configure migrate
    Old config
    -----------------------------------------------------
    ~/.ecs/config
@@ -129,7 +136,7 @@ Assume project name is "myApp"
 ### Existing customer second run
 1. Force migrating config, already knows what is going to happen
    ```
-   $ ecs-cli configure --force-migrate
+   $ ecs-cli configure migrate --force
    [WARN] Please read the following changes carefully: <link to documentation>
    - compose-project-name-prefix and compose-service-name-prefix is deprecated, you can continue to specify your desired names using the runtime flag --project-name
    - cfn-stack-name-prefix no longer exists, if you wish to continue to use your existing CFN stack, please specify the full stack name, otherwise it will be defaulted to    amazon-ecs-cli-setup-<cluster_name>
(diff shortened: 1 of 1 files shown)


proposals/remove-prefix-from-config.md
```diff
@@ -0,0 +1,180 @@
...
+or in the "license" file accompanying this file. This file is distributed
+on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+express or implied. See the License for the specific language governing
+permissions and limitations under the License.
+-->
+
+# Introduction
+As the CLI moves to multiple cluster configurations, the prefixes in the old configuration will be removed because:
+1. Bad user experience and not really a cluster configuration.
+1. Every new customers need to understand what the prefixes are for in order to know the names of the resources created by ECS CLI
```
markriggins: grammar:     Every new customer needs to

proposals/remove-prefix-from-config.md
```diff
@@ -0,0 +1,180 @@
...
+# Solution
+When users run `ecs-cli configure`, the prefixes will be removed from the config.
+```yaml
+default: my_config
+clusters:
+  my_config:
+  - cluster: demo
+  - region: us-west-2
+```
+
```
markriggins: The terms "config" and "cluster" are conflated.

This config YAML file supports multiple "configurations", but the YAML refers to them as multiple "clusters", using the key **clusters:**.    The **default:** vlaue of my_config shows that we are choosing a default **configuration** not a default **cluster**.

If these are cluster-configurations, then we should call them **cluster-configurations:**
```
default: my_config
cluster-configurations:
   my_config:
   - cluster: demo
   - region: us-west-2
```

proposals/remove-prefix-from-config.md
```diff
@@ -0,0 +1,180 @@
...
+   compose-project-name-prefix = ecscompose-
+   compose-service-name-prefix = ecscompose-service-
+   cfn-stack-name-prefix       = amazon-ecs-cli-setup-
+
+   New configs
+   -----------------------------------------------------
+   ~/.ecs/cluster.config
+   default: ecs
+   clusters:
+     ecs:
```
markriggins: recommend using **cluster-configurations:** instead of **clusters:**.     Otherwise the key below **ecs:** is a cluster?  That's going to confuse a lot of people.

markriggins: the filename **cluster.config** nourishes the idea that we are configuring a **single** cluster, instead of a set of cluster configurations.    Suggest:  **cluster-configurations.yml** or **cluster.configs** or **clusters.config**

proposals/remove-prefix-from-config.md
markriggins: If we dropped the notion of "cluster configurations"  and instead focused entirely on the clusters themselves.   It would it be possible to shorten this to 

```
default: demo
clusters:
  demo:
  - region: us-west-2
 production:
 - region: us-west-2
```
What value does the name "my_config" add if "my_config" has a singular value for "cluster?"
This would also make more sense with the notion of configuring multiple clusters, instead having multiple configurations.

proposals/remove-prefix-from-config.md
yinshiua: Thanks for the suggestions, naming is something that I am still working on.

**clusters:** as the key in the yaml is confusing, many people has raised this concern. I am currently considering **cluster-configs**, **cluster-configurations**, or just simply **configs**.

I like the file naming as **clusters.config**

proposals/remove-prefix-from-config.md
yinshiua: I think it's a good idea to simplify this to use cluster name as the configuration name, however, this would become a problem if someone wish to configure the same cluster in a different region.

markriggins:
if its in a different region, with different instances, is it really the
same cluster?

Mark Riggins

Cloud COMPUTING ARCHITECTURE AND SECurity ConSULTANT

336.312.7125

www.SocialCode.com <http://socialcode.com/>

yinshiua:
@markriggins It is possible that you want to use the same cluster name in a different region right?

markriggins:
That's true, since clusters are within regions.   But taking that into account, I would expect the yaml config file to take a similar approach:

```
regions: 
  us-west-2:
    default: demo
    clusters:
      demo:
      production:
  us-west-1:
    default: demo
    clusters:
      demo:
      production:
```

otherwise we wind up having to introduce names for our various configurations for no other benefit than differentiating between clusters with the same name.

From a disaster-recovery, and scaling perspective, it makes a `lot of sense` to have symetric clusters in multiple regions.

yinshiua:
@markriggins Thanks for the suggestion, but we're trying to introduce the concept of cluster configurations. It would be unclear what's the name of the configuration that the user can specify. 

After giving some more thoughts to it, I am still more inclined to your [other suggestion](https://github.com/aws/amazon-ecs-cli/pull/252/files#r121167647) of renaming "clusters:" to "cluster-configurations:"

Issue #97, opened 2016-04-06
Error listing tasks, InvalidParameterException: ownerId longer than 36


bryceray1121:
`ecs-cli compose -p "production_prod-ecscpu1_alerts" -f "/data/application/workspaces/production_prod-ecscpu1_alerts/artisan-run-alerts.yml"`

> RuntimeError: b'time="2016-04-06T20:53:58Z" level=error msg="Error listing tasks" error="InvalidParameterException: ownerId longer than 36.\n\tstatus code: 400, request id: af2c9d53-fc39-11e5-8d9c-871efba49583" request={\n  Cluster: "prod-ecscpu1",\n  DesiredStatus: "RUNNING",\n  StartedBy: "ecscompose-production_prod-ecscpu1_alerts"\n} \ntime="2016-04-06T20:53:58Z" level=fatal msg="InvalidParameterException: ownerId longer than 36.\n\tstatus code: 400, request id: af2c9d53-fc39-11e5-8d9c-871efba49583" \n',

Can't seem to get around this error and appears to be no documentation on what ownerId might be.

uttarasridhar:
ecs-cli uses the ["started-by"](http://docs.aws.amazon.com/cli/latest/reference/ecs/list-tasks.html) field to determine (and list) the tasks started by this compose project.
By default, it generates the string for started-by as "ecscompose-projectname", in this case that would be "ecscompose-production_prod-ecscpu1_alerts", which has 41 characters > 36 words.

With the [v0.3.0](https://github.com/aws/amazon-ecs-cli/releases/tag/v0.3.0) release, we have made the prefix for "started-by" configurable by supplying **`--compose-project-name-prefix`** to `configure` command.

**Example:** 
You can specify an empty string (--compose-project-name-prefix "") with this option to omit the default prefix ("ecscompose-"). 
```
`ecs-cli configure --region us-west-2 --access-key $AWS_ACCESS_KEY_ID --secret-key $AWS_SECRET_ACCESS_KEY --cluster ecs-cli --compose-project-name-prefix ""`
```
By setting it to empty, all the **successive** tasks started by ecs-cli compose would have started-by field as "project-name", which is "production_prod-ecscpu1_alerts"

You can find more information in the [docs](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-configure.html).

Hope that helps.

AlexTes:
[...]

PR #258, opened 2017-06-12
Use Task group to filter on list tasks instead of StartedBy

yinshiua:
closes https://github.com/aws/amazon-ecs-cli/issues/97

# Testing
```
$ make build

$ make test
```
Tasks:
```
$ ecs-cli compose up

$ ecs-cli compose ps
// Shows the correct task IDs (does not include the ones started by service)
```
Service:
```
$ ecs-cli compose service up

$ ecs-cli compose service ps
// Shows the correct task IDs
```


ecs-cli/modules/cli/compose/entity/entity_helper.go
```diff
@@ -120,7 +120,18 @@ func CollectTasksWithStatus(entity ProjectEntity, status string, filterLocal boo
 	result := []*ecs.Task{}
 
 	err := entity.Context().ECSClient.GetTasksPages(request, func(respTasks []*ecs.Task) error {
-		result = append(result, respTasks...)
+		// Filter the results by task.Group
+		if entity.EntityType() == "task" && filterLocal {
```
aaithal: 1. Can you use a const for comparison instead of `"task"`? Also, it'd make more sense for `EntityType()` to return an enum instead of a string (https://github.com/aws/amazon-ecs-cli/blob/master/ecs-cli/modules/cli/compose/entity/entity.go#L42)

2. Can you update the comment for `CollectTasksWithStatus` to reflect this change as well? Right now, `filterLocal` is only mentioned in the context of `startedBy`

ecs-cli/modules/cli/compose/entity/entityType/entityType.go
```diff
@@ -0,0 +1,23 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package entityType
```
aaithal: this can just be `type`. `enitity.entityType` stutters.

yinshiua:
yes, I've tested using running `ecs-cli compose up` with the old cli and run `ecs-cli compose ps` with the change.

ecs-cli/modules/cli/compose/entity/entityType/entityType.go
yinshiua: **type** is used in go, don't think I can name it as type, but I can name it as types. What do you think? Right now is entityType.Type, but with the change it'll be types.EntityType

aaithal: `types` is fine. You can just name it `types.Type` ?

PR #278, opened 2017-07-14
The cli config file is now formatted as yaml instead of ini

PettitWesley:
## Notes
- This feature is not to be used by itself; it is one step in the development of the multiple cluster configs proposal (#251) 
- Three new files, config_v1.go, readwriter_v1.go, and readwriter_test_v1.go were created and some code was reorganized in the files (moved between files)- unfortunately, this makes the diff look a bit odd it some places.

## Features
- new yaml config is stored as config (overwrites old config)
- fully backwards compatible; old ini formatted config file can still be read
- This change will allow us to implement #251 

## Testing 
- Reading old config file still works
- Configure command now writes to the yaml formatted config
- Reading new yaml formatted config works

```
$ make build
```

```
$ make test
```


ecs-cli/modules/config/config_v1.go
```diff
@@ -0,0 +1,142 @@
...
+//   a) AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY (OR)
+//   b) AWS_ACCESS_KEY and AWS_SECRET_KEY
+//  2) ECS Config - attempts to fetch the credentials from the ECS config file
+//  3) AWS Profile - attempts to use credentials (aws_access_key_id, aws_secret_access_key) or assume_role (role_arn, source_profile) from AWS profile name
+//    a) profile name from ECS config file (OR)
+//    b) AWS_PROFILE environment variable (OR)
+//    c) AWS_DEFAULT_PROFILE environment variable (defaults to 'default')
+//  4) EC2 Instance role
+func (cfg *CliConfig) ToAWSSession() (*session.Session, error) {
+	svcConfig := aws.Config{}
```
aaithal: Why is this not being doing in `toAWSSessionWithConfig`? Also, why not name this `NewAWSSession` or something similar? I am trying to understand why there are 2 methods here in place of having just one.

ecs-cli/modules/config/readwriter.go
```diff
@@ -52,12 +47,8 @@ type IniReadWriter struct {
...
 }
 
-// NewReadWriter creates a new Parser object.
-func NewReadWriter() (*IniReadWriter, error) {
-	dest, err := newDefaultDestination()
-	if err != nil {
-		return nil, err
-	}
+// NewIniReadWriter creates a new Ini Parser object for the old ini configs
+func NewIniReadWriter(dest *Destination) (*IniReadWriter, error) {
```
aaithal: Should probably be named `NewINIReadWriter`

ecs-cli/modules/config/readwriter.go
```diff
@@ -68,14 +59,53 @@ func NewReadWriter() (*IniReadWriter, error) {
 }
 
 // GetConfig gets the ecs-cli config object from the config file.
-func (rdwr *IniReadWriter) GetConfig() (*CliConfig, error) {
-	to := new(CliConfig)
-	err := rdwr.cfg.MapTo(to)
+func (rdwr *IniReadWriter) GetConfig() (*CliConfig, map[interface{}]interface{}, error) {
```
aaithal: Why is the second return type `map[interface{}]interface{}`? Can it contain more specific types? Can you add some documentation detailing what's being returned from this method?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,149 @@
...
+// ecs-cli config. Sample ecs-cli config:
+// cluster: test
+// aws_profile:
+// region: us-west-2
+// aws_access_key_id:
+// aws_secret_access_key:
+// compose-project-name-prefix: ecscompose-
+// compose-service-name-prefix:
+// cfn-stack-name-prefix: ecs-cli-
+type YamlReadWriter struct {
```
aaithal: Should be `YAMLReadWriter`

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,149 @@
...
+	// Warn the user if ini path also exists
+	iniPath := iniConfigPath(rdwr.destination)
+	if _, iniErr := os.Stat(iniPath); iniErr == nil {
+		logrus.Warnf("Writing yaml formatted config to %s/.ecs/%s.\nIni formatted config still exists in %s/.ecs/%s.", os.Getenv("HOME"), yamlConfigFileName, os.Getenv("HOME"), iniConfigFileName)
+	}
+
+	// If config file exists, set permissions first, because we may be writing creds.
+	if _, err := os.Stat(path); err == nil {
+		if err = os.Chmod(path, configFileMode); err != nil {
+			logrus.Errorf("Unable to chmod %s to mode %s", path, configFileMode)
```
aaithal: Can you log the error as well?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,149 @@
...
+			return err
+		}
+	}
+
+	data, err := yaml.Marshal(cliConfig)
+	if err != nil {
+		return err
+	}
+	if err = ioutil.WriteFile(path, data, configFileMode.Perm()); err != nil {
+		logrus.Errorf("Unable to write config to %s", path)
```
aaithal: Same here. Can you log the error as well?

ecs-cli/modules/config/config_v1.go
PettitWesley: Good point- I'll refactor that. 

(Technically I did not write this code tho actually- As I stated in the about, it was suggested to me that I rearrange the code- so config.go became config_v1.go and includes everything that is used now- whereas config.go is the old stuff that is being kept around for backwards compatibility. Unfortunately this makes it look like I have changed a lot more code than I actually have. I think in retrospect I should not have changed the file names. The blame suggests this code was written 2 years ago by Uttara. My guess is that may be it was done this way to facilitate testing?)

aaithal: I understand. Unless there are compelling reasons to retain the internal method, i'd recommend refactoring it so that we can reduce some loc's.

ecs-cli/modules/config/config.go
```diff
@@ -13,31 +18,19 @@
...
+// iniCLIConfig is the struct used to map to the ini config.
+// This is to allow us to read old ini based config files
+// CliConfig has been updated to use the yaml annotations
+type iniCLIConfig struct {
+	*iniSectionKeys `ini:"ecs"`
 }
 
-// SectionKeys is the struct embedded in CliConfig. It groups all the keys in the 'ecs' section in the ini file.
-type SectionKeys struct {
+// SectionKeys is the struct embedded in oldCliConfig. It groups all the keys in the 'ecs' section in the ini file.
```
aaithal: nit: `oldCliConfig` -> `iniCLIConfig`

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,155 @@
...
+	"path/filepath"
+
+	yaml "gopkg.in/yaml.v2"
+
+	"github.com/Sirupsen/logrus"
+)
+
+const (
+	yamlConfigFileName = "config.yml"
+	configFileMode     = os.FileMode(0600)
```
aaithal: nit: it'd be good to document what permissions this translates to as you might have to extend CLI to work on non Linux platform in the future.

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,155 @@
...
+	configMap := make(map[interface{}]interface{})
+	// read the raw bytes of the config file
+	iniPath := iniConfigPath(rdwr.destination)
+	yamlPath := yamlConfigPath(rdwr.destination)
+
+	_, iniErr := os.Stat(iniPath)
+	_, yamlErr := os.Stat(yamlPath)
+	if yamlErr == nil {
+		if err := readYAML(yamlPath, configMap, to); err != nil {
+			return nil, nil, err
```
aaithal: Is it worth logging why it's an error here?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	} else if iniErr == nil { // file exists
+		// old ini config
+		iniReadWriter, err := NewINIReadWriter(rdwr.destination)
+		if err != nil {
+			return nil, nil, err
```
aaithal: Is it worth logging why it's an error here?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		to, configMap, err = iniReadWriter.GetConfig()
+		if err != nil {
+			return nil, nil, err
```
aaithal: Is it worth logging why it's an error here?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+	} else {
+		// if neither file existed we return the yaml error
+		return nil, nil, yamlErr
```
aaithal: Is it worth logging why it's an error here?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,155 @@
...
+	iniPath := iniConfigPath(rdwr.destination)
+	if _, iniErr := os.Stat(iniPath); iniErr == nil {
+		logrus.Warnf("Writing yaml formatted config to %s/.ecs/%s.\nIni formatted config still exists in %s/.ecs/%s.", os.Getenv("HOME"), yamlConfigFileName, os.Getenv("HOME"), iniConfigFileName)
+	}
+
+	// If config file exists, set permissions first, because we may be writing creds.
+	if _, err := os.Stat(path); err == nil {
+		if err = os.Chmod(path, configFileMode); err != nil {
+			logrus.Errorf("Unable to chmod %s to mode %s", path, configFileMode)
+			logrus.Error(err.Error())
```
aaithal: You can combine this with the previous error line itself. Also, with `logrus`, you're discouraged to use `Errorf`m but instead do `WithFields`. Please replace all `Errorf`, `Info` etc with `WithFields` method.

ecs-cli/modules/config/readwriter_v1.go
PettitWesley: I've now added error logging inside the readYAML function itself.

PettitWesley: I've now added error logging in readYAML() itself

ecs-cli/modules/config/readwriter_v1.go
PettitWesley: So there are only 2 functions in readwriter_v1.go that are used elsewhere in the CLI- Save and GetConfig. These functions are only called in one place each, and each time there is a statement after the call that logs the error (if there was one). Therefore I am thinking of just removing all error code from the file entirely- what do you think?

ecs-cli/modules/config/readwriter.go
```diff
@@ -46,42 +41,79 @@ type ReadWriter interface {
...
+	oldFormat := &iniCLIConfig{iniSectionKeys: new(iniSectionKeys)}
+	err := rdwr.cfg.MapTo(oldFormat)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return to, nil
+	// Create the configMap
+	if rdwr.IsKeyPresent(ecsSectionKey, "cluster") {
```
yinshiua: Should we only check if the prefixes exist? The old format expect the other fields to be always present, so we can assume that the key always exists in this case.

ecs-cli/modules/config/readwriter.go
```diff
@@ -95,65 +127,13 @@ func (rdwr *IniReadWriter) IsInitialized() (bool, error) {
...
-}
-
-func newIniConfig(dest *Destination) (*ini.File, error) {
+func newINIConfig(dest *Destination) (*ini.File, error) {
 	iniCfg := ini.Empty()
-	path := configPath(dest)
-	logrus.Debugf("using config file: %s", path)
+	path := iniConfigPath(dest)
 	if _, err := os.Stat(path); err != nil {
 		// TODO: handle os.isnotexist(path) and other errors differently
```
yinshiua: We should remove this and the debug log. If the ini path doesn't exist, we now throw the yaml error.

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+const (
+	yamlConfigFileName = "config.yml"
```
yinshiua: Can we use the same file name "config"?

ecs-cli/modules/config/readwriter_v1.go
yinshiua: NewCliParams logs the error. We will be double logging the same information, but I do think we should add debug logs in the readYaml method to help understand which stage of yaml parsing failed.

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,143 @@
...
+	configMap := make(map[interface{}]interface{})
+	// read the raw bytes of the config file
+	iniPath := iniConfigPath(rdwr.destination)
+	yamlPath := yamlConfigPath(rdwr.destination)
+
+	_, iniErr := os.Stat(iniPath)
+	_, yamlErr := os.Stat(yamlPath)
+	if yamlErr == nil {
+		if err := readYAML(yamlPath, configMap, to); err != nil {
+			return nil, nil, err
```
yinshiua: Can do `return to, configMap, err` here?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+	} else if iniErr == nil { // file exists
+		// old ini config
+		iniReadWriter, err := NewINIReadWriter(rdwr.destination)
+		if err != nil {
+			return nil, nil, err
+		}
+		to, configMap, err = iniReadWriter.GetConfig()
```
yinshiua: same here, can we just return all here?

ecs-cli/modules/config/readwriter_v1.go
PettitWesley: I can- but why would I? If there has been an error reading the YAML file, then the function that called GetConfig should check and respond to the error only. If there is a problem because the config or map is nil- then that means the calling code is wrong.

PettitWesley: On second thought, I realized that doing so could help me reduce the number of lines of code, and number of if statements, which would be nice. I think I'll do it.

ecs-cli/modules/config/readwriter.go
PettitWesley: I can remove this if you want- however this code will be deleted when the next pull request happens for the rest of the multiple cluster config. 

(So the reason I did it this way was so that the behavior of the map if the file is ini is the same as if it was yaml. If the file is yaml and cluster is somehow not present, then there will be no key for it in the map. Basically the contract for the GetConfig function would state that map contains the keys present in the config file (regardless of whether or not we have an intent to use them in the application).
Obviously, either way the code works, it won't matter because if the cluster is not present then eventually an error will be thrown.)

ecs-cli/modules/config/params.go
```diff
@@ -50,22 +51,24 @@ func recursiveFlagSearch(context *cli.Context, flag string) string {
...
-// NewCliParams creates a new ECSParams object from the config file.
-func NewCliParams(context *cli.Context, rdwr ReadWriter) (*CliParams, error) {
-	ecsConfig, err := rdwr.GetConfig()
+// NewCLIParams creates a new ECSParams object from the config file.
+func NewCLIParams(context *cli.Context, rdwr ReadWriter) (*CLIParams, error) {
+	ecsConfig, configMap, err := rdwr.GetConfig()
+
 	if err != nil {
-		logrus.Error("Error loading config: ", err)
+		errors.Wrap(err, "Error loading config")
```
yinshiua: This is not used at all. I think we can just return the error.Wrap here since this will be logged again during the execution.

ecs-cli/modules/config/readwriter.go
```diff
@@ -95,73 +121,17 @@ func (rdwr *IniReadWriter) IsInitialized() (bool, error) {
...
 	path := configPath(dest)
-	logrus.Debugf("using config file: %s", path)
-	if _, err := os.Stat(path); err != nil {
-		// TODO: handle os.isnotexist(path) and other errors differently
-		// error reading config file, create empty config ini.
-		logrus.Debugf("no config files found, initializing empty ini")
-	} else {
+	if _, err := os.Stat(path); err == nil {
 		err = iniCfg.Append(path)
 		if err != nil {
```
yinshiua: nit: can you combine line 133 and 132?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return nil
+}
+
+// GetConfig gets the ecs-cli config object from the config file.
+// map contains the keys that are present in the config file (maps string field name to string field value)
+// map is type map[interface{}]interface{} to ensure fowards compatibility with changes that will
+// cause certain keys to be mapped to maps of keys
+func (rdwr *YAMLReadWriter) GetConfig() (*CLIConfig, map[interface{}]interface{}, error) {
```
yinshiua: This is much easier to read 👍

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,144 @@
...
+	if err := os.MkdirAll(rdwr.destination.Path, *destMode); err != nil {
+		return errors.Wrapf(err, "Could not make directory %s", rdwr.destination.Path)
+	}
+	// set version
+	cliConfig.Version = configVersion
+
+	path := configPath(rdwr.destination)
+
+	// If config file exists, set permissions first, because we may be writing creds.
+	if _, err := os.Stat(path); err == nil {
```
yinshiua: Is this check needed?

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -0,0 +1,144 @@
...
+		if err = os.Chmod(path, configFileMode); err != nil {
+			return errors.Wrapf(err, "Could not set file permissions, %s, for path %s", configFileMode, path)
+		}
+	}
+
+	data, err := yaml.Marshal(cliConfig)
+	if err != nil {
+		return errors.Wrap(err, "Could not convert config to YAML")
```
yinshiua: What do you think of changing this to "Unable to parse the config"?

ecs-cli/modules/config/readwriter_v1_test.go
```diff
@@ -0,0 +1,341 @@
...
+	assert.Empty(t, readConfig.ComposeProjectNamePrefix, "Compose project prefix name should be empty.")
+	_, ok = configMap[composeServiceNamePrefixKey]
+	assert.True(t, ok, "Compose service name prefix should exist in config")
+	assert.Empty(t, readConfig.ComposeServiceNamePrefix, "Compose service prefix name should be empty.")
+	_, ok = configMap[composeProjectNamePrefixKey]
+	assert.True(t, ok, "Compose project name prefix should exist in config")
+	assert.Empty(t, readConfig.CFNStackNamePrefix, "CFNStackNamePrefix should be empty.")
+}
+
+func TestConfigFileTruncation(t *testing.T) {
```
yinshiua: I don't think this test is needed since we are not writing config in ini formats.

ecs-cli/modules/config/readwriter_v1_test.go
```diff
@@ -0,0 +1,341 @@
...
+	assert.Empty(t, readConfig.ComposeProjectNamePrefix, "Compose project prefix name should be empty.")
+	_, ok = configMap[composeServiceNamePrefixKey]
+	assert.True(t, ok, "Compose service name prefix should exist in config.")
+	assert.Empty(t, readConfig.ComposeServiceNamePrefix, "Compose service prefix name should be empty.")
+	_, ok = configMap[cfnStackNamePrefixKey]
+	assert.True(t, ok, "CFNStackNamePrefix should exist in config.")
+	assert.Empty(t, readConfig.CFNStackNamePrefix, "CFNStackNamePrefix should be empty.")
+}
+
+func TestPrefixesDefaultOldINIFormat(t *testing.T) {
```
yinshiua: This test is not much different from `TestPrefixesEmptyOldINIFormat`

ecs-cli/modules/config/readwriter_v1_test.go
```diff
@@ -0,0 +1,341 @@
...
+	assert.NoError(t, err, "Error reading config")
+	_, ok := configMap[cfnStackNamePrefixKey]
+	assert.False(t, ok, "CFNStackNamePrefix should not exist in config")
+	_, ok = configMap[composeServiceNamePrefixKey]
+	assert.False(t, ok, "Compose service name prefix should not exist in config")
+	_, ok = configMap[composeProjectNamePrefixKey]
+	assert.False(t, ok, "Compose project name prefix should not exist in config")
+}
+
+func TestPrefixesDefaultNewYAMLFormat(t *testing.T) {
```
yinshiua: This test is not much different from `TestPrefixesEmptyNewYAMLFormat`

ecs-cli/modules/config/readwriter_v1.go
PettitWesley: It is. TestConfigPermissions will fail if it is removed.

ecs-cli/modules/config/readwriter_v1_test.go
PettitWesley: Good catch- this is a mistake on my part. These test cases are supposed to test different things.

ecs-cli/modules/config/readwriter_v1_test.go
PettitWesley: Good catch- this is a mistake on my part. These test cases are supposed to test different things.

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -122,6 +122,9 @@ func (rdwr *YAMLReadWriter) Save(cliConfig *CLIConfig) error {
 	path := configPath(rdwr.destination)
 
 	// If config file exists, set permissions first, because we may be writing creds.
+  // This is necessary because ioutil.WriteFile only sets the permissions
+  // if the file is being created for the first time; this handles the case
+  // where the file already exists
```
yinshiua: nit: indent

ecs-cli/modules/config/readwriter_v1.go
```diff
@@ -130,8 +133,9 @@ func (rdwr *YAMLReadWriter) Save(cliConfig *CLIConfig) error {
 
 	data, err := yaml.Marshal(cliConfig)
 	if err != nil {
-		return errors.Wrap(err, "Could not convert config to YAML")
+		return errors.Wrap(err, "Unable to parse the config")
 	}
+  // WriteFile will not change the permissions of an existing file
```
yinshiua: nit: looks like this needs to be indented. not sure if it's a github formatting issue

ecs-cli/modules/config/params.go
```diff
@@ -57,7 +56,6 @@ func NewCLIParams(context *cli.Context, rdwr ReadWriter) (*CLIParams, error) {
 
 	if err != nil {
 		errors.Wrap(err, "Error loading config")
-		logrus.Error(err)
 		return nil, err
```
yinshiua: `return errors.Wrap(err, "Error loading config")`

ecs-cli/modules/config/readwriter_v1_test.go
yinshiua: Should we change this to yaml?

PR #302, opened 2017-08-24
Simple Integration tests

### LINKED ISSUE #306: ecs-cli returns successful exit codes even when an error occurs
I'm running version 0.6.4 (5b0fd9c). See:

```
$ ecs-cli up
ERRO[0000] Error executing 'up': You must either specify a custom role with the '--instance-role' flag or set the '--capability-iam' flag 
$ echo $?
0
$ ecs-cli --invalid-arg compose service up
Incorrect Usage. flag provided but not defined: -invalid-arg
(...)
$ echo $?
0
```

Some commands return the expected error exit code, like:

```
$ ecs-cli foobar
No help topic for 'foobar'
$ echo $?
3
```

A similar issue was reported on https://github.com/aws/amazon-ecs-cli/issues/116.

PettitWesley:
closes #306 

## Features
- Script runs a series of ECS CLI commands on an ec2 instance. 
- This script does not provide full integration tests- it only checks the exit code of the command (it additionally also scans the output of the commands for the words "error" or "fatal" and then displays any lines with those words to the user). I think of it as a "sanity check" for any code change.
- The raw output of each command is stored in `~/test_output.txt`.
- Set up information is logged to `~/test_log.txt`
- Pulls code from a given Github username, and a given branch (ie. its meant to test code on your fork of the ECS CLI).
- Runs tests on a provided ec2 instance.
- It prints results in color!!!

## Example
```
./scripts/test_commands.sh -c IntegrationTestCluster -r us-east-2
-a $AWS_ACCESS_KEY_ID -s $AWS_SECRET_ACCESS_KEY -k MyFirstKeyPair
-p ~/Downloads/KeyPair.pem -i ec2-18-220-215-243.us-east-2.compute.amazonaws.com
-u PettitWesley -b master
```
Failures:
screen shot 2017-08-24 at 2 04 40 pm

Success:
screen shot 2017-08-29 at 5 31 52 pm

### DIFF
--- a/scripts/run_commands.sh
+++ b/scripts/run_commands.sh
@@ -37,40 +37,27 @@ expect_success_no_log() {
 	fi
 }
 
-# install git and go
-echo "TYPE y|yes then enter to proceed."
-sudo yum install git go >> ~/ecs-cli-test-results/test_log.txt
-# have to respond yes to prompt
-# get CLI
-export GOPATH="$HOME/go"
-go get github.com/aws/amazon-ecs-cli >> ~/ecs-cli-test-results/test_log.txt
-cd $GOPATH/src/github.com/aws/amazon-ecs-cli
-url="https://github.com/"
-url+=$gitname
-url+="/amazon-ecs-cli.git"
-git remote add fork $url &>> ~/ecs-cli-test-results/test_log.txt
-git fetch fork &>> ~/ecs-cli-test-results/test_log.txt
-git checkout "fork/${branch}"
-make build
-
-rm ~/ecs-cli-test-results/test_output.txt # clean up from past tests
+rm ~/ecs-cli-test-results/test_output.txt # clean up from past tests (in case this instance has been used before)
+# create the results directory and file; in case this instance is being used for the first time
+mkdir ~/ecs-cli-test-results/
+touch ~/ecs-cli-test-results/test_output.txt
 
 # test commands
 
 # configure
-expect_success_no_log ./bin/local/ecs-cli configure --region $region --access-key $access --secret-key $secret --cluster $cluster
+expect_success_no_log ./ecs-cli configure --region $region --access-key $access --secret-key $secret --cluster $cluster
 # up
-expect_success ./bin/local/ecs-cli up --capability-iam --keypair $keypair --size 1 --instance-type t2.medium --force
+expect_success ./ecs-cli up --capability-iam --keypair $keypair --size 1 --instance-type t2.medium --force
 # create a service
-expect_success ./bin/local/ecs-cli compose --file ./integration-tests/docker-compose.yml service up
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml service up
 # take down service
-expect_success ./bin/local/ecs-cli compose --file ./integration-tests/docker-compose.yml service down
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml service down
 # create a task
-expect_success ./bin/local/ecs-cli compose --file ./integration-tests/docker-compose.yml up
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml up
 # take down task
-expect_success ./bin/local/ecs-cli compose --file ./integration-tests/docker-compose.yml down
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml down
 # take down cluster
-expect_success ./bin/local/ecs-cli down --force
+expect_success ./ecs-cli down --force
 
 
 # Sanity Check- search output file for error messages
--- a/scripts/run_commands.sh
+++ b/scripts/run_commands.sh
@@ -42,6 +42,28 @@ rm ~/ecs-cli-test-results/test_output.txt # clean up from past tests (in case th
 mkdir ~/ecs-cli-test-results/
 touch ~/ecs-cli-test-results/test_output.txt
 
+if ! [ -z "${gitname}" ]; then
+	# Not testing local changes, testing a branch instead
+	echo "Testing $branch"
+	# install git and go
+	echo "TYPE y|yes then enter to proceed."
+	sudo yum install git go >> ~/ecs-cli-test-results/test_log.txt
+	# have to respond yes to prompt
+	# get CLI
+	export GOPATH="$HOME/go"
+	go get github.com/aws/amazon-ecs-cli >> ~/ecs-cli-test-results/test_log.txt
+	cd $GOPATH/src/github.com/aws/amazon-ecs-cli
+	url="https://github.com/"
+	url+=$gitname
+	url+="/amazon-ecs-cli.git"
+	git remote add fork $url &>> ~/ecs-cli-test-results/test_log.txt
+	git fetch fork &>> ~/ecs-cli-test-results/test_log.txt
+	git checkout "fork/${branch}"
+	make build
+
+	cd bin/local/
+fi
+
 # test commands
 
 # configure
--- a/scripts/run_commands.sh
+++ b/scripts/run_commands.sh
@@ -13,8 +13,19 @@ LRED='\033[0;31m'
 LGREEN='\033[0;32m'
 NC='\033[0m' # No Color
 
+diff_binaries() {
+	if diff ./ecs-cli ./ecs-cli-"${1}"; then
+		echo -e "${GREEN}SUCCESS: ecs-cli-latest and ecs-cli-${1} are the same${NC}"
+	else
+		echo "-------"
+		echo -e "${RED}FAILED: ecs-cli-latest and ecs-cli-${1} differ.${NC}"
+		echo "-------"
+		FAILURES+=1
+	fi
+}
+
 expect_success() {
-	if "${@}" &>> $TEST_RESULT_DIR/test_output.txt; then
+	if "${@}" >> $TEST_RESULT_DIR/test_output.txt 2>&1; then
 		echo -e "${GREEN}SUCCEEDED${NC}: ${LGREEN}${@}${NC}"
 	else
 		echo "-------"
@@ -26,7 +37,7 @@ expect_success() {
 
 # used for commands that whose arguments might contain sensitive information
 expect_success_no_log() {
-	if "${@}" &>> $TEST_RESULT_DIR/test_output.txt; then
+	if "${@}" >> $TEST_RESULT_DIR/test_output.txt 2>&1; then
 		echo -e "${GREEN}SUCCEEDED${NC}: ${LGREEN}${1} ${2}${NC}"
 	else
 		echo "-------"
@@ -52,8 +63,8 @@ if ! [ -z "${gitname}" ]; then
 	url="https://github.com/"
 	url+=$gitname
 	url+="/amazon-ecs-cli.git"
-	git remote add fork $url &>> $TEST_RESULT_DIR/test_log.txt
-	git fetch fork &>> $TEST_RESULT_DIR/test_log.txt
+	git remote add fork $url 2>&1 $TEST_RESULT_DIR/test_log.txt
+	git fetch fork 2>&1 $TEST_RESULT_DIR/test_log.txt
 	git checkout "fork/${branch}"
 	make build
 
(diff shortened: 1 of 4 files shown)


output
```diff
@@ -0,0 +1,21 @@
+TERM_PROGRAM=Apple_Terminal
```
samuelkarp: You probably didn't mean to check this file in.  Would be good to add it to .gitignore if you expect it to be produced on each run.

PettitWesley: Oh oops. That was created while I was developing. It isn't created each time; I'll delete it.

samuelkarp:
I'd recommend squashing these commits together so that the unintended file gets removed from git history.

ecs-cli/modules/cli/compose/compose_app.go
```diff
@@ -142,11 +141,3 @@ func ProjectDown(p ecscompose.Project, c *cli.Context) {
 		log.Fatal(err)
 	}
 }
-
-// common function for all unsupported operations
-func unsupportedOperation(c *cli.Context) {
```
yinshiua: Why is this removed?

PettitWesley: Because its not used at all. So I just removed it. `make build` and `make test` run fine without it.

scripts/test_commands.sh
```diff
@@ -0,0 +1,86 @@
...
+			;;
+	esac
+done
+
+if [ -z "${instance_url}" ]; then
+	usage
+	exit 1
+fi
+
+scp -i $keypath $(dirname "${0}")/run_commands.sh "ec2-user@${instance_url}":~/
```
SoManyHs: Could we use `rsync` here?

SoManyHs: (in case, for example, someone has an existing instance that they are unfrugally keeping up for testing, and have already copied the test script there at some point. Rsync in this scenario would be a tad faster, i believe)

scripts/test_commands.sh
```diff
@@ -0,0 +1,86 @@
+# EXAMPLE:
+# ./scripts/sanity_check.sh -c IntegrationTestCluster -r us-east-2
```
SoManyHs: Hmm, am I missing something? I don't see a `sanity_check.sh` file anywhere in the `/scripts` directory.

scripts/run_commands.sh
```diff
@@ -0,0 +1,70 @@
...
+# install git and go
+echo "TYPE y|yes then enter to proceed."
+sudo yum install git go >> ~/test_log.txt
+# have to respond yes to prompt
+# get CLI
+export GOPATH="$HOME/go"
+go get github.com/aws/amazon-ecs-cli >> ~/test_log.txt
+cd go/src/github.com/aws/amazon-ecs-cli
+url="https://github.com/"
+url+=$gitname
```
SoManyHs: We talked about this a bit off line, but just noting here that on first past, it's not immediately obvious where this and other env variables are being set. Perhaps worth adding a comment or some other kind of documentation on how the test files are supposed to fit together?

scripts/test_commands.sh
PettitWesley: oops, I changed the name

scripts/test_commands.sh
PettitWesley: That's true, `rsync` would probably be slightly faster in some cases. But I think `scp` is fine enough.

scripts/run_commands.sh
PettitWesley: Sure, I'll add a comment at the top of each file.

scripts/run_commands.sh
```diff
@@ -0,0 +1,92 @@
...
+	else
+		echo "-------"
+		echo -e "${RED}FAILED${NC}: ${LRED}${1} ${2}${NC}"
+		echo "-------"
+		FAILURES+=1
+	fi
+}
+
+# install git and go
+echo "TYPE y|yes then enter to proceed."
```
yinshiua: Can we move the comments here?

"Please make sure you have git and go installed. Do you wish to continue? [y/n]"

scripts/run_commands.sh
```diff
@@ -0,0 +1,92 @@
...
+sudo yum install git go >> ~/test_log.txt
+# have to respond yes to prompt
+# get CLI
+export GOPATH="$HOME/go"
```
yinshiua: Would this change my actual `GOPATH` if I'm running this on my laptop?

Can we only use it for this specific command similar to `Makefile`?

scripts/run_commands.sh
```diff
@@ -0,0 +1,92 @@
...
+go get github.com/aws/amazon-ecs-cli >> ~/test_log.txt
+cd go/src/github.com/aws/amazon-ecs-cli
```
yinshiua: We are assuming you're in GOPATH directory, can we make this `$GOPATH/src/...`

scripts/test_commands.sh
```diff
@@ -0,0 +1,91 @@
...
+
+if [ -z "${instance_url}" ]; then
+	usage
+	exit 1
+fi
+
+scp -i $keypath $(dirname "${0}")/run_commands.sh "ec2-user@${instance_url}":~/
+# exit 1
+# ARGS: cluster c, region r, access a, secret s, keypair k , keypath p, instance_url i, gitname u, branch b
+ssh -i $keypath "ec2-user@${instance_url}" "chmod +x run_commands.sh"
```
yinshiua: Do I need to provide my key pair path?

scripts/run_commands.sh
```diff
@@ -0,0 +1,92 @@
...
+url="https://github.com/"
+url+=$gitname
+url+="/amazon-ecs-cli.git"
+git remote add fork $url &>> ~/test_log.txt
+git fetch fork &>> ~/test_log.txt
```
yinshiua: We need a better home for these temp files, should we move this to ~/tmp?

scripts/run_commands.sh
```diff
@@ -0,0 +1,92 @@
...
+git checkout "fork/${branch}"
+make build
```
yinshiua: Is setting up git and go path really needed? Can you run make build first on your laptop and copy the binary to the EC2 instance?

You would run the following in order:
1. `make build` --> you will have a binary built already.
2. `make test` --> make sure unit tests works.
3. `make integration-test` --> copy the binary and the script to EC2 instance, and run all commands

scripts/test_commands.sh
PettitWesley: Yes

scripts/run_commands.sh
PettitWesley: This runs on the instance, which is presumably a clean instance without any of those installed on it. 

I don't understand what you mean be "Can we move the comments here"

scripts/run_commands.sh
PettitWesley: So no one should be running this script on their own laptop- its meant to run on an EC2 instance running Amazon Linux. `test_commands.sh` is what is run on your laptop.

scripts/run_commands.sh
PettitWesley: That's a good idea. I think I'd prefer to name it `~/test-results/` or something like that.

scripts/run_commands.sh
yinshiua: If I didn't read the script, I would not know by proceeding, go and git will be installed.

I thought the script assumes you already have an EC2 instance, it doesn't spin up an EC2 instance for you. I can use the same instance to run the integration test multiple times. Is it really necessary to install go and git on the EC2 instance? Why can't we just copy the binary needed for testing? With the setup today, I must check in my code and must of a forked repository in order for this to work.

scripts/run_commands.sh
PettitWesley: I considered this idea myself. It seems really really nice, but as I was implementing it I realized it will not work! See, the CLI supports both Mac and Linux- and the binaries are system dependent. So for example, if I use a Mac and then try to copy my built binary to an instance- it will not run on that instance, because it is built for the wrong platform.

scripts/run_commands.sh
yinshiua: to be more specific, what do you think of using `ecs-cli-test-results`

scripts/run_commands.sh
yinshiua: Ahh, I see. Can we use the binary from `make docker-build`? It will build two binaries, one for Mac and one for Linux.

scripts/run_commands.sh
PettitWesley: > Why can't we just copy the binary needed for testing?

I'm looking into it.

> With the setup today, I must check in my code and must of a forked repository in order for this to work.

Actually it can also test branches on the main repo, by setting the GitHub username to be "aws".

(I do understand that the script is not ideal as it functions now)

scripts/run_commands.sh
```diff
@@ -0,0 +1,101 @@
...
+rm ~/ecs-cli-test-results/test_output.txt # clean up from past tests (in case this instance has been used before)
+# create the results directory and file; in case this instance is being used for the first time
+mkdir ~/ecs-cli-test-results/
+touch ~/ecs-cli-test-results/test_output.txt
+
+if ! [ -z "${gitname}" ]; then
+	# Not testing local changes, testing a branch instead
+	echo "Testing $branch"
+	# install git and go
+	echo "TYPE y|yes then enter to proceed."
```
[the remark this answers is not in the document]
yinshiua: I can't seem to find where we are reading the answer of the prompt.

scripts/run_commands.sh
```diff
@@ -0,0 +1,101 @@
...
+	sudo yum install git go >> ~/ecs-cli-test-results/test_log.txt
+	# have to respond yes to prompt
```
yinshiua: Should there be some code that are doing what you described here?

scripts/run_commands.sh
```diff
@@ -0,0 +1,101 @@
...
+# create a service
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml service up
+# take down service
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml service down
+# create a task
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml up
+# take down task
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml down
+# take down cluster
+expect_success ./ecs-cli down --force
```
yinshiua: This is a good start. We would want to add other tests such as `compose service ps`.

scripts/run_commands.sh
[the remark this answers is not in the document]
PettitWesley: The shell reads the user input itself. Specifically, the yum install command accepts the user input. The only reason I put in this print statement is so that the user knows that they have to type yes. Otherwise the install will not proceed (normally yum install will print a prompt to the terminal, but in the script I'm capturing the output and saving it to a file (so that extraneous output does swamp the user's terminal).

scripts/run_commands.sh
PettitWesley: Good point. I'll add tests for those. Any others that you think I should add for v1 of this script? (As time goes on I think we can add more complicated tests, but for now I want to make sure it at least tests all the basic functionality of the CLI).

yinshiua: I think it would be good to have all the compose commands `start`,  `scale`, `create`, `run` and `stop` since they are commonly used and they all share some code, one Task Definition change might affect multiple commands. The other cluster related commands or ECR related commands can be done later.

scripts/run_commands.sh
```diff
@@ -0,0 +1,102 @@
...
+
+# test commands
+
+# configure
+expect_success_no_log ./ecs-cli configure --region $region --access-key $access --secret-key $secret --cluster $cluster
+# up
+expect_success ./ecs-cli up --capability-iam --keypair $keypair --size 1 --instance-type t2.medium --force
+# create a service
+expect_success ./ecs-cli compose --file ./integration-tests/docker-compose.yml service up
+# ps on th service
```
yinshiua: typo "the"

scripts/run_commands.sh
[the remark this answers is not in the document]
samuelkarp: You can use `yum -y install` to skip the prompt.  There's also the `yes` command that can be used for these kinds of things; you could also do `yes | yum install`.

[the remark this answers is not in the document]
PettitWesley: Thanks Sam!!

SoManyHs:
Based on team discussion, we're going to either use an internal test package or travis for this. Closing.

PR #331, opened 2017-09-25
Add support for cap_add and cap_drop

PettitWesley:
[On Friday, ECS added support for adding and dropping linux capabilities](https://aws.amazon.com/about-aws/whats-new/2017/09/amazon-ecs-adds-support-for-adding-or-dropping-linux-capabilities-to-containers/)

- added support for cap_add and cap_drop using the docker compose fields
- updated aws-sdk-go to the latest release

Testing Done

- Ran `ecs-cli compose up` with this docker compose file:

```
version: '2'
services:
  mysql:
    image: mysql
    cpu_shares: 100
    mem_limit: 524288000
    environment:
      MYSQL_ROOT_PASSWORD: password1234catscatscatsdogsdogspudding
    cap_add:
      - ALL
    cap_drop:
      - NET_ADMIN
      - SYS_ADMIN
```

Which resulted in the following Container Definition:
```
"containerDefinitions": [
    {
      "linuxParameters": {
        "capabilities": {
          "add": [
            "ALL"
          ],
          "drop": [
            "NET_ADMIN",
            "SYS_ADMIN"
          ]
        }
      },
```

### DIFF
--- a/ecs-cli/modules/utils/compose/convert_task_definition.go
+++ b/ecs-cli/modules/utils/compose/convert_task_definition.go
@@ -44,7 +44,7 @@ var supportedComposeYamlOptions = []string{
 	"cpu_shares", "command", "dns", "dns_search", "entrypoint", "env_file",
 	"environment", "extra_hosts", "hostname", "image", "labels", "links",
 	"logging", "log_driver", "log_opt", "mem_limit", "mem_reservation", "ports", "privileged", "read_only",
-	"security_opt", "ulimits", "user", "volumes", "volumes_from", "working_dir",
+	"security_opt", "ulimits", "user", "volumes", "volumes_from", "working_dir", "cap_add", "cap_drop",
 }
 
 var supportedComposeYamlOptionsMap = getSupportedComposeYamlOptionsMap()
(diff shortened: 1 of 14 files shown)


ecs-cli/modules/utils/compose/convert_task_definition.go
```diff
@@ -263,6 +263,14 @@ func convertToContainerDef(context *project.Context, inputCfg *config.ServiceCon
 		outputContDef.WorkingDirectory = aws.String(inputCfg.WorkingDir)
 	}
 
+	outputContDef.LinuxParameters = &ecs.LinuxParameters{Capabilities: &ecs.KernelCapabilities{}}
+	if inputCfg.CapAdd != nil {
```
yinshiua: is it possible that this will be empty instead of nil?

PettitWesley: Nope. Here is the line in libcompose for this field: 
```CapAdd          []string             yaml:"cap_add,omitempty"```

Which means that it will be nil if its not present. If on the other hand you specify it but don't give any capabilities (empty list), then you get this error:
```
ERRO[0000] Could not parse config for project test2 : Service 'mysql' configuration key 'cap_add' contains an invalid type, it should be an array.
```

PR #364, opened 2017-11-06
Add support for configuring multiple named ECS Profiles and cluster configurations

PettitWesley:
- Switched configuration format from INI to YAML
- Named Cluster Configurations are stored in `~/.ecs/config`
- Named ECS Profiles for Credentials are stored in `~/.ecs/credentials`
- Updated the order of resolution for credentials and region


kaihendry:
I can't seem to find any documentation or tools to help customers move from INI to YAML.

What am I missing?

PettitWesley:
@kaihendry, please see: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_CLI_migrating_config_files.html

You can use `ecs-cli configure migrate` to migrate an INI Config file to the new YAML format.

kaihendry:
Thanks, guess the most important thing in my migration was aligning the `--compose-service-name-prefix` so that services already configured with a load balancer will be seamless.

PR #438, opened 2018-03-01
Fail with an informative error message if cluster isn't set

petderek:
The CLI help indicates the cluster and region flags are optional.
However, this is only true if defaults are stored in .ecs/*

This commit checks for the presence of the 'cluster' flag, and
throws an error if not set. This way, we won't call the ECS API
and we present a clearer failure message.

#### Current (makes network call, unclear error)
```
➜  ecs-cli scale --capability-iam --size 0
FATA[0000] Error executing 'scale': InvalidParameterException: Clusters can not be blank.
	status code: 400, request id: b636f430-2ef5-11e8-bfed-5775413ad506
➜  ecs-cli ps
FATA[0000] Error executing 'ps InvalidParameterException: Clusters can not be blank.
	status code: 400, request id: 8e9625e2-2ef5-11e8-be00-2f76bba63d2b
```
#### New behavior (fails fast, reuses error from ecs-cli up)

```
➜  ecs-cli scale --capability-iam --size 0
FATA[0000] Error executing 'scale': Please configure a cluster using the configure command or the '--cluster' flag
➜  ecs-cli ps
FATA[0000] Error executing 'ps': Please configure a cluster using the configure command or the '--cluster' flag
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ bin/
 *.orig
 ecs-cli/vendor/pkg
 .vscode/*
+.idea/*
--- a/ecs-cli/modules/cli/cluster/cluster_app.go
+++ b/ecs-cli/modules/cli/cluster/cluster_app.go
@@ -115,13 +115,13 @@ func ClusterScale(c *cli.Context) {
 func ClusterPS(c *cli.Context) {
 	rdwr, err := config.NewReadWriter()
 	if err != nil {
-		logrus.Fatal("Error executing 'ps ", err)
+		logrus.Fatal("Error executing 'ps': ", err)
 	}
 
 	ecsClient := ecsclient.NewECSClient()
 	infoSet, err := clusterPS(c, rdwr, ecsClient)
 	if err != nil {
-		logrus.Fatal("Error executing 'ps ", err)
+		logrus.Fatal("Error executing 'ps': ", err)
 	}
 	os.Stdout.WriteString(infoSet.String(container.ContainerInfoColumns, displayTitle))
 }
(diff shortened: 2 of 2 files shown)


ecs-cli/modules/config/params.go
```diff
@@ -108,6 +109,15 @@ func NewCLIParams(context *cli.Context, rdwr ReadWriter) (*CLIParams, error) {
 		ecsConfig.CFNStackName = flags.CFNStackNamePrefixDefaultValue + ecsConfig.Cluster
 	}
 
+	if ecsConfig.Cluster == "" {
```
petderek: To be clear, it will only log a warning if cluster isn't specified anywhere -- including as a flag. 

If this check should be in a different place, let me know!

PettitWesley: So the issue with this is that it will log this warning for all commands, even the ECR commands, which do not need or have any use for a cluster name.

petderek: I'll move it over to 'ecs-cli ps'. I'll do a quick scan to make sure other codes work the same way.

ecs-cli/modules/config/params.go
```diff
@@ -108,6 +109,15 @@ func NewCLIParams(context *cli.Context, rdwr ReadWriter) (*CLIParams, error) {
...
+		logrus.Warn("No default cluster is set. You may want to run `ecs-cli configure`.")
+	}
+
+	if ecsConfig.Region == "" {
```
petderek: Should this stay?

allisaurus: will this still log if they've specified a "--region" flag? If not, then makes sense to me to leave it. Any calls w/o a region are going to fail

petderek: Turns out, its a little more complicated than I thought -- so I'm going to drop this.

The ecs-cli will fall back to the aws-cli's region config, which means at the best case we'd be asking for redundant configuration settings
```
# .ecs region unset, .aws region set
➜  ~ ecs ps --cluster default
WARN[0000] No default region is set. You may want to run `ecs-cli configure`. 
```

However, this code path isn't executed if both regions are unset. (It fails fast)
```
# .ecs and .aws region unset
➜  ~ ecs ps --cluster default
ERRO[0000] this be wehre da really bad happen           
FATA[0000] Error executing 'ps Set a region using ecs-cli configure command with the --region flag or AWS_REGION environment variable or --profile flag
➜  ~ ecs pull myrepo
FATA[0000] Error executing 'pull': Set a region using ecs-cli configure command with the --region flag or AWS_REGION environment variable or --profile flag
```

petderek:
I believe I have made the merge robots happy

PR #523, opened 2018-05-25
Fix bug in ps command; resolves #522

### LINKED ISSUE #522: ecs-cli ps fails for Fargate tasks that are provisioning
### Description

When a Fargate task is being launched, it goes through a "Provisioning" phase where the network interface is attached. The ECS CLI uses the network interface to display the IP address in the `ps` command. However, if the task is provisioning and there is not yet a network interface, it falls back to the container instance IP. This works for EC2 launch type tasks, but it doesn't work for Fargate, and causes a confusing/incorrect error message.

```
$ ecs-cli ps --cluster-config default                                       
FATA[0000] Error executing 'ps': No container instances for found tasks 
```

PettitWesley:
- ps command used to try to fall back to container instance IP if no network interfaces are found, this fails when a Fargate task is provisioning.
- With the fix, when 'awsvpc' tasks are provisioning, the 'ports' collumn will not show an IP address, it will simply show the ports.


allisaurus:
can you add some test output that shows the new (fixed) results of running PS while FG tasks are provisioning?

ecs-cli/modules/cli/compose/entity/entity_test.go
```diff
@@ -120,6 +121,43 @@ func taskDefinition() *ecs.TaskDefinition {
 	}
 }
 
+// Test case ensures that GetContainersForTasksWithTaskNetworking() can be given
+// a list of tasks with task networking but without ENIs attached (i.e. task is provisioning) and return correctly.
+func TestGetContainersForTasksWithTaskNetworkingENIProvisioning(t *testing.T) {
+	ecsTasks := []*ecs.Task{
```
allisaurus: can you add a comment here re: what this task is lacking? (i.e., "set up task without ENI/IP")

PettitWesley: I can, but I think the comment at the top of the test is pretty descriptive, its tasks with task networking, but without an ENIs (because the ENIs are still provisioning). I'm happy to edit that comment or add another one if you feel its needed, but I think that what I have there currently is sufficient and adding more would be redundant.

PettitWesley:
@allisaurus 

I ran ps repeatedly in one window while I was spinning up a fargate task in another. Before the ENI is attached, it shows this:

```
$ ecs-cli ps 
Name                                            State                Ports                 TaskDefinition
67d0af69-6828-4e91-b42c-9a78705fec7a/wordpress  PENDING              :80->80/tcp           compose3:1
```

I'm gonna change the `ps` printing code so that the `:` is removed. But I still want to show the ports, because the task does have ports defined, the IP just isn't available yet. So it will simply show `80->80/tcp`. Thoughts?

ecs-cli/modules/cli/compose/entity/entity_test.go
allisaurus: I guess from looking at this code and not being super familiar with it, it wasn't immediately apparent to me where the lack of ENI was supposed to be determined (in the task struct or in some resource/call made with it later), so a comment here would help me understand where we're setting up the ENI-less-ness (?) we're testing against.

allisaurus:
re: the Ports - agreed we should leave the ports, but instead of just leaving it out entirely would it be better to print (IP provisioning) or something like that instead? Would explain the lack if users are looking for it.

PettitWesley:
@allisaurus I realized that task networking tasks also should not have their IP shown when they are stopped (the network interface can be described for some time after it has been detached from the task, and its private IP was shown by the `ps` command). So I added a commit to show the IP as empty. 

Here's some testing output which shows the life cycle of an `awsvpc` task as shown by the fixed `ps` command:

Task Starting/Network Interface Provisioning:
```
$ ecs-cli ps --cluster-config compose3
Name                                            State                Ports                 TaskDefinition
d1e66a7b-9ae2-4fbe-a561-6123698cca4d/wordpress  PENDING              (PENDING):80->80/tcp  compose3:1
c90c0d3e-f08d-42e1-9d89-8a0fbd625b07/wordpress  STOPPED ExitCode: 0  80->80/tcp            compose3:1
```
Pending Task with network interface attached (Fargate task has not been allocated public IP yet):
```
$ ecs-cli ps --cluster-config compose3
Name                                            State                Ports                  TaskDefinition
d1e66a7b-9ae2-4fbe-a561-6123698cca4d/wordpress  PENDING              10.0.0.240:80->80/tcp  compose3:1
c90c0d3e-f08d-42e1-9d89-8a0fbd625b07/wordpress  STOPPED ExitCode: 0  80->80/tcp             compose3:1
```
Running:
```
$ ecs-cli ps --cluster-config compose3
Name                                            State                Ports                     TaskDefinition
d1e66a7b-9ae2-4fbe-a561-6123698cca4d/wordpress  RUNNING              34.241.203.17:80->80/tcp  compose3:1
c90c0d3e-f08d-42e1-9d89-8a0fbd625b07/wordpress  STOPPED ExitCode: 0  80->80/tcp                compose3:1
```
Stopped:
```
$ ecs-cli ps --cluster-config compose3
Name                                            State                Ports       TaskDefinition
c90c0d3e-f08d-42e1-9d89-8a0fbd625b07/wordpress  STOPPED ExitCode: 0  80->80/tcp  compose3:1
d1e66a7b-9ae2-4fbe-a561-6123698cca4d/wordpress  STOPPED ExitCode: 0  80->80/tcp  compose3:1
```

PR #526, opened 2018-05-29
Move logging unsupported fields into logger package

allisaurus:
Opened in place of https://github.com/aws/amazon-ecs-cli/pull/524 (rebased on latest)


ecs-cli/modules/cli/compose/logger/logger.go
```diff
@@ -0,0 +1,129 @@
...
+	"user",
+	"volumes", // v2
+	"volumes_from",
+	"working_dir",
+}
+
+var supportedComposeYamlOptionsMap = getSupportedComposeYamlOptionsMap()
+
+// Create set of supported YAML fields in docker compose v2 ServiceConfigs
+func getSupportedComposeYamlOptionsMap() map[string]bool {
```
PettitWesley: nit: I've always wondered why we don't use a map literal for this

PR #604, opened 2018-09-12
Move resource prefix to utils package

allisaurus:
Pull "amazon-ecs-cli-setup-" prefix out as a constant, reference where previously used as string value.

--

### DIFF
--- a/ecs-cli/modules/utils/utils.go
+++ b/ecs-cli/modules/utils/utils.go
@@ -19,6 +19,11 @@ import (
 	"os"
 )
 
+const (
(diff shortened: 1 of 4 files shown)


ecs-cli/modules/utils/utils.go
```diff
@@ -19,6 +19,11 @@ import (
 	"os"
 )
 
+const (
+	// EcsCliResourcePrefix is prepended to the names of resources created through the ecs-cli
+	EcsCliResourcePrefix = "amazon-ecs-cli-setup-"
```
PettitWesley: My eyes don't like this... this should be `ECSCLIResourcePrefix` to fit Golang style for initialisms.

PR #622, opened 2018-10-02
Ensure CFN Stack Names fit API length and charset requirements

PettitWesley:

### DIFF
--- a/ecs-cli/modules/cli/servicediscovery/servicediscovery_helper.go
+++ b/ecs-cli/modules/cli/servicediscovery/servicediscovery_helper.go
@@ -15,6 +15,7 @@ package servicediscovery
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 
 	"github.com/aws/amazon-ecs-cli/ecs-cli/modules/clients/aws/cloudformation"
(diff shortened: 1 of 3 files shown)


ecs-cli/modules/cli/servicediscovery/servicediscovery_helper.go
```diff
@@ -213,7 +214,23 @@ func getOutputIDFromStack(cfnClient cloudformation.CloudformationClient, stackNa
...
+// and truncating at 128 characters
+func sanitizeAndTruncate(s string) string {
+	reg, err := regexp.Compile("[^a-zA-Z0-9-]+")
+	if err != nil {
+		// the regex compiles, the unit tests verify this, there's no need to return this error
+		logrus.Fatal(err)
+	}
+	sanitized := reg.ReplaceAllString(s, "0")
+	if len(sanitized) > 128 {
+		return sanitized[:128]
```
allisaurus: I'm concerned that if we truncate the complete generated name (vs. cluster & service name individually), customers who have a very long cluster name can only ever have one SD stack in that cluster. Any stack name generated for another service in the same cluster will end up with the same generated name b/c the service name will be cut off. 

Can we instead cap the cluster & service components of the stack name to 40 char (or whatever the math ends up being)? This is similar to how the ECS console generates a new target group name for services with an LB (although the size limit there is much tighter)

ecs-cli/modules/cli/servicediscovery/servicediscovery_helper.go
```diff
@@ -213,7 +218,28 @@ func getOutputIDFromStack(cfnClient cloudformation.CloudformationClient, stackNa
 }
 
 func cfnStackName(stackName, cluster, service string) string {
```
allisaurus: can we rename the param `stackName` to `stackNameFmt` or something similar to be clear it's a format string (and not, say, something with more significance)?

allisaurus: (also apologies for not noticing earlier)

PR #650, opened 2018-10-24
Add README content for using private reg creds

allisaurus:

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Line Interface](http://aws.amazon.com/cli/) product detail page.
 		- [Using Route53 Service Discovery](#using-route53-service-discovery)
 	- [Viewing Running Tasks](#viewing-running-tasks)
 	- [Viewing Container Logs](#viewing-container-logs)
+    - [Using Private Registry Authentication](#using-private-registry-authentication)
 - [Amazon ECS CLI Commands](#amazon-ecs-cli-commands)
 - [Contributing to the CLI](#contributing-to-the-cli)
 - [License](#license)
@@ -839,16 +840,16 @@ OPTIONS:
 
 ## Using Private Registry Authentication
 
-If you want to use privately hosted container images with ECS, the ECS CLI can store your private registry credentials for you with AWS Secrets Manager and create an IAM role which ECS can use to access the credentials and private images. This allows you to:
+If you want to use privately hosted container images with ECS, the ECS CLI can store your private registry credentials in AWS Secrets Manager and create an IAM role which ECS can use to access the credentials and private images. This allows you to:
 
-* Store private registry credentials within AWS for use with ECS or other services
+* Store private registry credentials within AWS for use with ECS
 * Add the permissions needed to use your registry secrets to a new or existing Task Execution Role
 * Automatically add your private registry credentials to your task definition when running a task or service
 
 Using privately hosted images with the ECS CLI is done in two parts:
 
 1) Create new AWS Secrets Manager secrets and an IAM Task Execution Role with `ecs-cli registry-creds up`
-2) Run `ecs-cli compose up` to create and run a task definition that includes the new resources
+2) Run `ecs-cli compose` commands to create and run a task definition that includes the new resources
 
 ### Storing private registry credentials with `ecs-cli registry-creds up`
 
@@ -856,13 +857,16 @@ To get started, first create an input file that contains the name of your regist
 
 ```
 # file name: cred_input.yml
+# when using environment variables, only '${VAR_NAME}' format is supported
+
 version: '1'
 registry_credentials:
   my-registry.example.com: 
-    username: myUserName
-    password: ${MY_PASSWORD}
-    kms_key_id: <optional KMS Key ID to use to encryt the new secret>
-    container_names:
+    secrets_manager_arn:        # required when using (with no modification) or updating an existing secret
+    username: myUserName        # required when creating or updating a new secret
+    password: ${MY_PASSWORD}    # required when creating or updating a new secret  
+    kms_key_id:                 # optional custom KMS Key ID to use to encrypt new secret
+    container_names:            # required to match credential resources with docker-compose services
       - web
       - log
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
 --timestamps, -t           [Optional] Shows timestamps on each line in the log output.
 ```
 
+## Using Private Registry Authentication
+
+If you want to use privately hosted container images with ECS, the ECS CLI can store your private registry credentials for you with AWS Secrets Manager and create an IAM role which ECS can use to access the credentials and private images. This allows you to:
```
PettitWesley: suggestion: "the ECS CLI can store your private registry credentials in AWS Secrets Manager"

README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
...
+
+* Store private registry credentials within AWS for use with ECS or other services
```
PettitWesley: Are we sure we can say "or other services"? The secret format that we use is specific to ECS, I am not sure there is any guarantee it will necessarily work for other services.

README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
...
+* Add the permissions needed to use your registry secrets to a new or existing Task Execution Role
+* Automatically add your private registry credentials to your task definition when running a task or service
+
+Using privately hosted images with the ECS CLI is done in two parts:
+
+1) Create new AWS Secrets Manager secrets and an IAM Task Execution Role with `ecs-cli registry-creds up`
+2) Run `ecs-cli compose up` to create and run a task definition that includes the new resources
```
PettitWesley: `ecs-cli compose create` is the command that just creates a task definition. But I'd recommend re-wording this to just reference "The compose commands"

README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
...
+
+To get started, first create an input file that contains the name of your registry and the credentials needed to access it:
+
+```
+# file name: cred_input.yml
+version: '1'
+registry_credentials:
+  my-registry.example.com: 
+    username: myUserName
+    password: ${MY_PASSWORD}
```
PettitWesley: `${}` is supported but just using $ isn't right? I'd call that out somewhere

README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
...
+    kms_key_id: <optional KMS Key ID to use to encryt the new secret>
+    container_names:
+      - web
+      - log
+```
+
+In this example, we're storing credentials for a registry called `my-registry.example.com` and passing in the password with an environment variable. `container_names` is a list of the `service_names` in your Docker Compose project which need access to images in this registry. If you don't plan to use the output of `registry-creds up` to launch a task or service with `compose`, then you can leave this field empty.
```
PettitWesley: suggestion: may be "with `compose` commands"?

README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
...
+To get started, first create an input file that contains the name of your registry and the credentials needed to access it:
+
+```
+# file name: cred_input.yml
+version: '1'
+registry_credentials:
+  my-registry.example.com: 
+    username: myUserName
+    password: ${MY_PASSWORD}
+    kms_key_id: <optional KMS Key ID to use to encryt the new secret>
```
PettitWesley: encryt

README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
...
+        awslogs-stream-prefix: privateImageApp
+```
+
+Now run the command `ecs-cli compose up` to launch a task. The ECS CLI will automatically detect and use the newest `ecs-registry-creds` file within the current directory:
+
+```
+$~\privateImageApp> ecs-cli compose up
+INFO[0000] Found ecs-registry-creds file C:\Users\myuser\regcreds\regCredTest\ecs-registry-creds_20181023T210805Z.yml
+INFO[0000] Using ecs-registry-creds value arn:aws:secretsmanager:region:aws_account_id:secret:amazon-ecs-cli-setup-my-registry.example.com-VeDqXm container name=web option name=credentials_parameter
+Using ecs-registry-creds value arn:aws:secretsmanager:region:aws_account_id:secret:amazon-ecs-cli-setup-my-registry.example.com-VeDqXm container name=log option name=credentials_parameter
```
PettitWesley: Sort of random, I should have noticed this in a previous PR, but I think its weird how it says "Using ecs-registry-creds value" and "Found ecs-registry-creds file". Why not replace "ecs-registry-creds" with "registry credentials". To me, "ecs-registry-creds" with the dashes makes it seem like that's a field or some other sort of technical thing which I should recognize, but its not right? An "ecs-registry-creds" isn't a thing right?

README.md
```diff
@@ -837,6 +837,137 @@ OPTIONS:
...
+INFO[0000] Using ECS task definition TaskDefinition="privateImageApp:1"
+INFO[0000] Starting container... container=bf35a813-dd76-4fe0-b5a2-c1334c2331f4/web
+INFO[0000] Starting container... container=bf35a813-dd76-4fe0-b5a2-c1334c2331f4/log
+INFO[0012] Describe ECS container status container=bf35a813-dd76-4fe0-b5a2-c1334c2331f4/web desiredStatus=RUNNING lastStatus=PENDING taskDefinition="privateImageApp:1"
+INFO[0013] Describe ECS container status container=bf35a813-dd76-4fe0-b5a2-c1334c2331f4/log desiredStatus=RUNNING lastStatus=PENDING taskDefinition="privateImageApp:1"
+INFO[0018] Started container... container=bf35a813-dd76-4fe0-b5a2-c1334c2331f4/web desiredStatus=RUNNING lastStatus=RUNNING taskDefinition="privateImageApp:1"
+INFO[0018] Started container... container=bf35a813-dd76-4fe0-b5a2-c1334c2331f4/log desiredStatus=RUNNING lastStatus=RUNNING taskDefinition="privateImageApp:1"
+```
+
+ The within your new task definition `privateImageApp:1`, the container definitions for both `web` and `log` should have your "my-registry.example.com" secret as a `credentialsParameter` and the `executionRoleArn` will be "myTaskExecutionRole"
```
PettitWesley: This should probably be two sentences, and should also end with a period.

README.md
allisaurus: I was following the [convention we use](https://github.com/aws/amazon-ecs-cli/blob/master/ecs-cli/modules/utils/compose/convert_task_definition.go#L281) to refer to values from `ecs-params.yml`. When we override docker-compose values in a task def with values from this file, we message the customer in a similar way: "Using ecs-params value as override" / "Using ecs-params value as override (was %v but is now %v)"

I'm not married to "ecs-registry-creds" as a term, but it is the (sans timestamp) name of the src file and I think it's helpful to indicate where the value came from. 

However, if we want to avoid creating a new convention for now I think omitting the term would be OK too b/c when those values are used we do identify the container & task def param (credentials_parameter) being updated.

PettitWesley: Ah I see, I guess I'm fine with leaving it the way it is

README.md
```diff
@@ -837,6 +838,141 @@ OPTIONS:
 --timestamps, -t           [Optional] Shows timestamps on each line in the log output.
 ```
 
+## Using Private Registry Authentication
+
+If you want to use privately hosted container images with ECS, the ECS CLI can store your private registry credentials in AWS Secrets Manager and create an IAM role which ECS can use to access the credentials and private images. This allows you to:
+
+* Store private registry credentials within AWS for use with ECS
+* Add the permissions needed to use your registry secrets to a new or existing Task Execution Role
```
PettitWesley: I think this should read  "Add the permissions needed to add your registry secrets to a new or existing Task Execution Role"

Issue #642, opened 2018-10-19
Allow --create-log-groups when only some containers use CW Logs


PettitWesley:
### Summary
The `--create-log-groups`requires that all containers are configured to use CloudWatch logs. It would be more useful if it worked when only some containers use CW Logs.

### Description
* Which specific command was run? 
```
$ ecs-cli compose service up --cluster-config kinesis-fargate --create-log-groups
INFO[0000] Using ECS task definition                     TaskDefinition="docker-amazon-kinesis-agent:6"
FATA[0000] Container 'apache' is not configured to use CloudWatch logs; logConfiguration ('logging' in Docker Compose) is a required container definition field 
```
* Which version of the CLI you are using? 
`v1.9.0`
* Which version of Go are you using? 
`go1.11.1`
* What platform are you using to run ECS CLI commands? (E.g. Linux, MacOSX, Windows)
OS X

### Config files
* docker-compose.yml

```
version: '2'
services:
  kinesis-agent:
    image: ************.dkr.ecr.us-east-2.amazonaws.com/amazon-kinesis-agent:latest
    volumes:
      - shared_volume:/tmp/aws-kinesis-agent/
    logging:
      driver: awslogs
      options:
        awslogs-group: tutorial
        awslogs-region: us-east-2
        awslogs-stream-prefix: kinesis-agent
  apache:
    image: **********dkr.ecr.us-east-2.amazonaws.com/apache:latest
    volumes:
      - shared_volume:/var/log/apache
volumes:
  shared_volume:
```

Log Groups are successfully created when `apache` is given a CW Logs configuration.

PR #694, opened 2019-01-04
Allow --create-log-groups when only some containers use awslogs

PettitWesley:
*Issue #, if available:*
#642

### DIFF
--- a/ecs-cli/modules/cli/logs/logs_app.go
+++ b/ecs-cli/modules/cli/logs/logs_app.go
@@ -309,6 +309,9 @@ func logConfigMisMatchError(taskDef *ecs.TaskDefinition, fieldName string) error
 // CreateLogGroups creates any needed log groups for the task definition to use CloudWatch Logs
 func CreateLogGroups(taskDef *ecs.TaskDefinition, logClientFactory cwlogsclient.LogClientFactory) error {
 	for _, container := range taskDef.ContainerDefinitions {
+		if container.LogConfiguration == nil || container.LogConfiguration.LogDriver == nil || aws.StringValue(container.LogConfiguration.LogDriver) != "awslogs" {
+			continue
+		}
 		logConfig, err := getContainerLogConfig(container)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


PettitWesley:
Testing:

```
version: '2'
services:
  container1:
    image: apache
    logging:
      driver: awslogs
      options:
        awslogs-group: somegroup
        awslogs-region: us-east-2
        awslogs-stream-prefix: somestream
  container2:
    image: apache
```

```
$ ecs-cli compose --file create-logs.yml service up --create-log-groups --cluster-config compose3
INFO[0000] Using ECS task definition                     TaskDefinition="compose3:12"
INFO[0001] Created Log Group somegroup in us-east-2     
INFO[0001] Updated the ECS service with a new task definition. Old containers will be stopped automatically, and replaced with new ones  desiredCount=1 force-deployment=false service=compose3
INFO[0001] Service status                                desiredCount=1 runningCount=1 serviceName=compose3
INFO[0064] (service compose3) has started 1 tasks: (task fa2e6bdd-1edd-4ae4-8d33-25fdf8821ada).  timestamp="2019-01-08 18:35:13 +0000 UTC"
```

ecs-cli/modules/cli/logs/logs_app.go
```diff
@@ -309,6 +309,9 @@ func logConfigMisMatchError(taskDef *ecs.TaskDefinition, fieldName string) error
 // CreateLogGroups creates any needed log groups for the task definition to use CloudWatch Logs
 func CreateLogGroups(taskDef *ecs.TaskDefinition, logClientFactory cwlogsclient.LogClientFactory) error {
 	for _, container := range taskDef.ContainerDefinitions {
+		if container.LogConfiguration == nil || container.LogConfiguration.LogDriver == nil || aws.StringValue(container.LogConfiguration.LogDriver) != "awslogs" {
+			continue
+		}
 		logConfig, err := getContainerLogConfig(container)
 		if err != nil {
 			return err
```
allisaurus: will this still fail if NO containers use awslogs? (and if so, can we add a test for it?)

PettitWesley: It won't fail if no containers use awslogs- in that case nothing happens. Normally if log groups are created, then you'll get an Info message about it, if no containers use awslogs, then you get no message. I guess using `--create-log-groups` when no containers use awslogs is a user error- but I don't really want that to cause the command to fail. What do you think?

allisaurus: I don't think it should fail either, though it feels like we may want to `WARN` in this case so customers know that option isn't doing anything. Would adding that check blow up the scope of the current change? If not let's add it, if so I don't consider it a blocker.

PR #748, opened 2019-03-11
Upgrade aws-sdk-go to v1.18.0

efekarakus:
*Description of changes:* Upgrade and [pin the version](https://golang.github.io/dep/docs/Gopkg.toml.html#version) of aws-sdk-go to 1.18.0


efekarakus:
* Updated with `dep ensure -update github.com/aws/aws-sdk-go`
* Ran `make docker-test` and tests pass 
* Ran `make build` and used the `compose` command to start two containers

ecs-cli/Gopkg.toml
```diff
@@ -25,7 +25,7 @@
 
 [[constraint]]
   name = "github.com/aws/aws-sdk-go"
-  version = "1.15.2"
+  version = "=1.18.0"
```
PettitWesley: Why is there an = sign?

efekarakus: Good question! @SoManyHs and I recently found out that if you don't have an '=' sign,then `dep ensure` will upgrade the version of the dependency up to the next major version. To pin a version number we should use '='.

> When you specify a version without an operator, dep automatically uses the ^ operator by default. 

https://golang.github.io/dep/docs/Gopkg.toml.html#version

PettitWesley: Interesting, good to know 👍

PR #765, opened 2019-04-23
Integ tests following the Fargate tutorial

efekarakus:
*Description of changes:* Add integration tests that follow our Fargate tutorial (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-cli-tutorial-fargate.html)


efekarakus:
- [x] `make integ-test` passes locally.
- [x] Dryrun on CodeBuild also passes:
```
Building ecs-cli... 
./scripts/build_binary.sh ./bin/local 
Running integration tests... 
go test -tags integ -v ./ecs-cli/integ/... 
?       github.com/aws/amazon-ecs-cli/ecs-cli/integ [no test files] 
?       github.com/aws/amazon-ecs-cli/ecs-cli/integ/cfn [no test files] 
?       github.com/aws/amazon-ecs-cli/ecs-cli/integ/cmd [no test files] 
=== RUN   TestCreateClusterWithFargateService 
--- PASS: TestCreateClusterWithFargateService (365.44s) 
    configure.go:42: Created config ecs-cli-canary-dec75407-a609-48b5-bff5-e583edee05ee-fargate-config 
    up.go:60: Created cluster ecs-cli-canary-dec75407-a609-48b5-bff5-e583edee05ee-fargate-cluster in stack amazon-ecs-cli-setup-ecs-cli-canary-dec75407-a609-48b5-bff5-e583edee05ee-fargate-cluster 
    compose.go:72: Created service with name e2e-fargate-test-service 
    compose.go:85: Project e2e-fargate-test-service has 1 running containers 
    compose.go:116: Scaled the service e2e-fargate-test-service to 2 tasks 
    compose.go:85: Project e2e-fargate-test-service has 2 running containers 
    compose.go:147: Deleted service e2e-fargate-test-service 
    cfn.go:58: Stack amazon-ecs-cli-setup-ecs-cli-canary-dec75407-a609-48b5-bff5-e583edee05ee-fargate-cluster does not exist as expected 
    down.go:51: Deleted stack amazon-ecs-cli-setup-ecs-cli-canary-dec75407-a609-48b5-bff5-e583edee05ee-fargate-cluster 
PASS 
ok      github.com/aws/amazon-ecs-cli/ecs-cli/integ/e2e 365.441s 
?       github.com/aws/amazon-ecs-cli/ecs-cli/integ/stdout  [no test files] 
  ```

File: ecs-cli/integ/cfn/cfn.go [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+func TestStackNameExists(t *testing.T, stackName string) {
+	client := newClient(t)
+	resp, err := client.DescribeStacks(&cloudformation.DescribeStacksInput{
+		StackName: aws.String(stackName),
+	})
+	if err != nil {
+		assert.FailNowf(t, "unexpected CloudFormation error during DescribeStacks", "wanted no errors, got %v", err)
+	}
+	if resp.Stacks == nil {
+		assert.FailNow(t, "stacks should not be nil")
```
petderek: FFTI... but you can delete some lines if you do this:

```
import "github.com/stretchr/testify/require"

require.NoErrorf(t, err, "message", var)

// similar functions for NotNil, Equal, etc
```

ecs-cli/integ/cfn/cfn.go
```diff
@@ -0,0 +1,78 @@
...
+func TestNoStackName(t *testing.T, stackName string) {
+	client := newClient(t)
+	resp, err := client.DescribeStacks(&cloudformation.DescribeStacksInput{
+		StackName: aws.String(stackName),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "ValidationError":
+				t.Logf("Stack %s does not exist as expected", stackName)
```
PettitWesley: nit: when I first read this statement I didn't realize it was saying that the test was successful. How about:

Success: Stack %s does not exist

Or something like that

ecs-cli/integ/stdout/stdout.go
```diff
@@ -13,18 +13,19 @@
...
+	"github.com/stretchr/testify/assert"
 )
 
 // Stdout is the standard output content from running a test.
 type Stdout []byte
 
-// HasAllSnippets returns true if stdout contains each snippet in wantedSnippets, false otherwise.
-func (b Stdout) HasAllSnippets(t *testing.T, wantedSnippets []string) bool {
+// TestHasAllSnippets returns true if stdout contains each snippet in wantedSnippets, false otherwise.
+func (b Stdout) TestHasAllSnippets(t *testing.T, wantedSnippets []string) bool {
```
PettitWesley: nit: `TestHasAllOutput`?

ecs-cli/integ/cmd/compose.go
```diff
@@ -0,0 +1,186 @@
...
+	cmd := integ.GetCommand(args)
+
+	// When
+	out, err := cmd.Output()
+	if err != nil {
+		assert.FailNowf(t, "Failed to create service", "Error %v running %v", err, args)
+	}
+
+	// Then
+	stdout.Stdout(out).TestHasAllSnippets(t, []string{
```
PettitWesley: Hmm. I kind of think that it would be better to call ecs:DescribeService to verify that the command succeeded, rather than checking the ECS CLI output.

PettitWesley: I'm fine with just validating the command output, but I'd prefer if in addition we also made an API call to verify/check.

ecs-cli/integ/cfn/cfn.go
efekarakus: Thanks for sharing this! It made the code so much neater ✨

ecs-cli/integ/stdout/stdout.go
efekarakus: renamed to `TestHasAllSubstrings`

ecs-cli/integ/cmd/compose.go
efekarakus: Yeah, so I also debated about this as well. I wrote the tests from the point of view of a user. How do they validate if `compose service scale` worked? they probably run `compose service ps`. That's why `fargate_service_test` is structured like:
```go
// Increase the number of running tasks
cmd.TestServiceScale(t, project, 2)
cmd.TestServicePs(t, project, 2)
```
In scenarios where it didn't seem satisfying, like in `TestUp` I use the `cfnClient` to validate that the stack's name is following our naming template. 

I couldn't think of some integration to test behind the scenes to validate that it's doing the right thing that PS doesn't do for us. I'm definitely open to suggestions here though :)

PettitWesley: Okay, that makes sense to me. This satisfies my concern :)

ecs-cli/integ/cmd/compose.go
```diff
@@ -0,0 +1,183 @@
...
+// TestServiceUp runs `ecs-cli compose service up` for a project.
+func TestServiceUp(t *testing.T, p *Project) {
+	// Given
+	args := []string{
+		"compose",
+		"--project-name",
+		p.Name,
+		"service",
+		"up",
+		"--create-log-groups",
```
allisaurus: What's the behavior if these log groups already exist? I believe it outputs a warning (vs. failing) - will this interfere with later `stdout` assertions?

allisaurus: Can you confirm the cmd behavior when the log groups already exist? My understanding is they'll be created the first time this is run, but creation is silently (or with a warning) skipped once they exist. 

It doesn't look like we're asserting on the existence or name of the log groups later in the test, so I'm wondering how useful this is to include (here and in the docker-compose.yml)?

ecs-cli/integ/e2e/fargate_service_test.go
```diff
@@ -0,0 +1,94 @@
...
+    image: wordpress
+    ports:
+      - "80:80"
+    logging:
+      driver: awslogs
+      options: 
+        awslogs-group: tutorial
+        awslogs-region: us-east-1
+        awslogs-stream-prefix: wordpress`
+	err := ioutil.WriteFile("./docker-compose.yml", []byte(content), os.ModePerm)
```
allisaurus: I don't see a where this file (or `ecs-params.yml`) are cleaned up. What if someone was running the tests on their laptop vs. in CodeBuild (which would terminate the instance after exiting)? I'd rather these not create resources that stick around after the test finishes.

Instead of creating a "real" file in the current directory, can we create a (temp) file in a [temp directory](https://golang.org/pkg/io/ioutil/#TempDir) and pass the name of that file into `compose` with `-f` ? If we end up running multiple tests that invoke compose, using+overwriting the same file could lead to weird failures if something get missed. It also means we can't run the e2e tests in parallel one day (assuming they'll live in the same dir).

ecs-cli/Gopkg.toml
```diff
@@ -65,7 +65,7 @@
 
 [[constraint]]
   name = "github.com/stretchr/testify"
-  version = "1.2.1"
+  version = "=1.2.1"
```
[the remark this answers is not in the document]
efekarakus: I pinned it down to a specific version number that we know that works, instead of the default that will use a higher patch version if it exists (see behavior in https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#version).
We previously ran into an issue before where running `dep ensure -update` on the aws-sdk-go resulted in a broken build 😝. By pinning it we know what we get.

Upgraded to "=1.3.0"

ecs-cli/integ/cmd/compose.go
efekarakus: You're right it skips it with a WARNING:
```
INFO[0000] Using ECS task definition                     TaskDefinition="efe-test:1"
WARN[0001] Failed to create log group tutorial in us-east-1: The specified log group already exists 
INFO[0001] Created an ECS service                        service=efe-test taskDefinition="efe-test:1"
INFO[0002] Updated ECS service successfully              desiredCount=1 force-deployment=false service=efe-test
```

Removed.

ecs-cli/integ/cmd/configure.go
```diff
@@ -0,0 +1,68 @@
...
+// Package cmd contains ECS CLI command tests.
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
```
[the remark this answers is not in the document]
efekarakus: This is what [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) formats it as for me :/ 

Kept it as is.

ecs-cli/integ/e2e/fargate_service_test.go
efekarakus: 👍 great suggestion. Done! The `ioutil.TempFile` function is pretty great.

ecs-cli/integ/cmd/configure.go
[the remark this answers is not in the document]
allisaurus: ¯\_(ツ)_/¯

ecs-cli/Gopkg.toml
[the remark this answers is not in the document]
allisaurus: Ah ha! Didn't realize that's what broke us on the aws-sdk-go build before.

Issue #772, opened 2019-05-15
Integration tests for ecs-cli local


efekarakus:
### Summary
Add integration tests for `ecs-cli local`. These tests should follow the steps defined in the Local Container Endpoints blog post: https://aws.amazon.com/blogs/compute/a-guide-to-locally-testing-containers-with-amazon-ecs-local-endpoints-and-docker-compose/

bvtujo:
Closing this after merging in #903 and #914. 
903: Addition of integration tests which actually launch and tear down local containers
914: Expansion of scope of integration tests, validation of docker-compose contents, validation of running containers using Docker SDK.

PR #775, opened 2019-05-20
Teardowns the local network on "local stop"

efekarakus:
**Description of changes**: Remove the local endpoints container and the local network if the user doesn't need them anymore. Freeing up resources on their machine.

----

**Testing**  
- [x] Unit tests passed
- [x] Link to issue or PR for the integration tests: https://github.com/aws/amazon-ecs-cli/issues/772

**Documentation**  
- [N/A] Updated our README
----


efekarakus:
## Manual Tests

#### 1. `local stop` when only the endpoints container is running
```shell
# Given
$ docker ps               
CONTAINER ID        IMAGE                                         COMMAND                  CREATED             STATUS              PORTS               NAMES
697ab62e760c        amazon/amazon-ecs-local-container-endpoints   "/local-container-en…"   55 seconds ago      Up 54 seconds       80/tcp              amazon-ecs-local-container-endpoints
$ docker network ls     
NETWORK ID          NAME                DRIVER              SCOPE
94ac9ec2ce2b        bridge              bridge              local
949837dcb884        ecs-local-network   bridge              local
3803595a49d9        host                host                local
59a17e60cd02        none                null                local

# When
$ dev-ecs-cli local stop                
INFO[0000] The network ecs-local-network has no more running tasks, stopping the endpoints containers... 
INFO[0000] Stopped the amazon-ecs-local-container-endpoints container successfully, removing it... 
INFO[0000] Removed the amazon-ecs-local-container-endpoints container successfully, removing the ecs-local-network network... 
INFO[0000] Removed the ecs-local-network network successfully 

# Then
$ docker ps                  
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
$ docker network ls     
NETWORK ID          NAME                DRIVER              SCOPE
94ac9ec2ce2b        bridge              bridge              local
3803595a49d9        host                host                local
59a17e60cd02        none                null                local
```

#### 2. `local stop` when a task is running
```shell
# Given
$ docker ps                 
CONTAINER ID        IMAGE                                         COMMAND                  CREATED              STATUS              PORTS               NAMES
01bd513e4d4f        ecs-local-other_app                           "/bin/sh -c '/bin/ba…"   10 seconds ago       Up 9 seconds                            ecs-local-other_app_1
03e24e741a3f        amazon/amazon-ecs-local-container-endpoints   "/local-container-en…"   About a minute ago   Up About a minute   80/tcp              amazon-ecs-local-container-endpoints

# When
$ dev-ecs-cli local stop  # does nothing

# Then
$ docker ps                      
CONTAINER ID        IMAGE                                         COMMAND                  CREATED              STATUS              PORTS               NAMES
01bd513e4d4f        ecs-local-other_app                           "/bin/sh -c '/bin/ba…"   About a minute ago   Up About a minute                       ecs-local-other_app_1
03e24e741a3f        amazon/amazon-ecs-local-container-endpoints   "/local-container-en…"   About a minute ago   Up About a minute   80/tcp              amazon-ecs-local-container-endpoints
```

#### 3. `local stop` when a task is running without local endpoints container
```shell
# Given
$ docker ps                  
CONTAINER ID        IMAGE                 COMMAND                  CREATED             STATUS              PORTS               NAMES
01bd513e4d4f        ecs-local-other_app   "/bin/sh -c '/bin/ba…"   2 minutes ago       Up 2 minutes                            ecs-local-other_app_1

# When
$ dev-ecs-cli local stop     
WARN[0000] The ecs-local-other_app_1 container is running in the ecs-local-network network without the amazon-ecs-local-container-endpoints container, please stop it first 

# Then
$ docker ps                      
CONTAINER ID        IMAGE                 COMMAND                  CREATED             STATUS              PORTS               NAMES
01bd513e4d4f        ecs-local-other_app   "/bin/sh -c '/bin/ba…"   3 minutes ago       Up 3 minutes                            ecs-local-other_app_1
```

#### 4. `local stop` with no running containers in the network
```shell
# Given

$ docker network ls
NETWORK ID          NAME                DRIVER              SCOPE
94ac9ec2ce2b        bridge              bridge              local
4cb93d5d5a88        ecs-local-network   bridge              local
3803595a49d9        host                host                local
59a17e60cd02        none                null                local
$ docker container ps --all       
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES

# When
$ dev-ecs-cli local stop       
INFO[0000] The network ecs-local-network has no more running tasks, stopping the endpoints containers... 
INFO[0000] Removed the ecs-local-network network successfully 

# Then
$ docker network ls       
NETWORK ID          NAME                DRIVER              SCOPE
94ac9ec2ce2b        bridge              bridge              local
3803595a49d9        host                host                local
59a17e60cd02        none                null                local
```

ecs-cli/modules/cli/local/network/teardown.go
```diff
@@ -0,0 +1,147 @@
...
+	defer cancel()
+
+	resp, err := d.NetworkInspect(ctx, EcsLocalNetworkName, types.NetworkInspectOptions{})
+	if err != nil {
+		logrus.Fatalf("Failed to inspect network %s due to %v", EcsLocalNetworkName, err)
+	}
+
+	if len(resp.Containers) > 1 {
+		// Has other containers running in the network
+		return true
```
nathanpeck: I'd suggest adding a log statement here so that users have more visibility in the case where teardown is being skipped because they had launched other containers in the network

ecs-cli/modules/cli/local/local_app.go
```diff
@@ -44,3 +44,14 @@ func Up(c *cli.Context) {
 	}
 	network.Setup(docker)
 }
+
+// Stop stops a running local ECS task.
+//
+// If the user stops the last running task in the local network then also remove the network.
+func Stop(c *cli.Context) {
+	docker, err := client.NewEnvClient() // Temporary client created to test network.Teardown()
```
PettitWesley: You might want to consider doing something like I did here: https://github.com/awslabs/amazon-ecs-local-container-endpoints/blob/master/local-container-endpoints/clients/docker/client.go#L49

I once or twice ran into weird problems where the client API version was too new for my local docker. If you lock it to the min version that is needed, then customers are safe.

ecs-cli/modules/cli/local/network/teardown.go
```diff
@@ -14,25 +14,134 @@
...
+	return false
+}
+
+func stopEndpointsContainer(d containerStopper) {
+	ctx, cancel := context.WithTimeout(context.Background(), dockerTimeout)
+	defer cancel()
+
+	err := d.ContainerStop(ctx, localEndpointsContainerName, nil)
+	if err != nil {
+		if strings.Contains(strings.ToLower(err.Error()), "no such container") {
```
PettitWesley: Why is this needed?

ecs-cli/modules/cli/local/network/teardown.go
efekarakus: Done! Added `"%d other task(s) running locally, skipping network removal."`

ecs-cli/modules/cli/local/network/teardown.go
efekarakus: Great question, looks like the comment below wasn't precise enough :P 

This error happens if the user stopped all their task containers with `docker stop` instead of `ecs-cli local stop`. In that situation, `hasRunningTasksInNetwork()` will return `false` and we will try to stop the endpoints container again. But since it was already stopped this error is returned. 

We don't want to `log.Fatal` while trying to stop an already stopped container, it should be a no-op hence this if-statement. 

I updated the comment to `// The containers in the network were already stopped by the user using "docker stop", do nothing.`

efekarakus: An alternative is to not invoke `stopEndpointsContainer()` or `removeEndpointsContainer()` if there are 0 containers in the local network.

However, I think that approach pushes complexity upwards instead of downwards (into the function) so I don't prefer it as much. Also if you run `ecs-cli local stop` on the same task concurrently (two terminals for example) then one of those could trigger this error and fail with fatal instead of continuing cleanly by not doing any work.

PettitWesley: Got it... my only concern is that this is yet another place where we are using string matching on errors returned by Docker :/

efekarakus: Yeah, I agree but I think Docker locked themselves and can't change their error messages otherwise they'll break everyone's code :P  

Here is a similar example from the agent: https://github.com/aws/amazon-ecs-agent/blob/24c71162368b3a9090689532e5c035fe2bccf133/agent/dockerclient/dockerapi/docker_client.go#L693

PettitWesley: This is my least favorite thing in Go...

ecs-cli/modules/cli/local/network/setup.go
```diff
@@ -77,7 +77,7 @@ const (
 )
 
 // Setup creates a user-defined bridge network with a running Local Container Endpoints container. If the network
-// already exists or the container is already running then the operation is a no-op.
+// already exists or the container is already running then this function does nothing.
```
PettitWesley: Very minor nit: this change and the other one in teardown.go probably shouldn't be in the "generate mocks" commit.

If you plan to squash all changes to one commit prior to merge, then don't worry about it

PettitWesley:
```
                                     # #  ( )
                                  ___#_#___|__
                              _  |____________|  _
                       _=====| | |            | | |==== _
                 =====| |.---------------------------. | |====
   <--------------------'   .  .  .  .  .  .  .  .   '--------------/
     \                                                             /
      \___________________________________________________________/
  wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww
wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww
   wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww
```

PR #812, opened 2019-06-26
Deleted duplicate const definition and added comment for const

iamhopaul123:
**Description of changes**:
Removed duplicate definition for default output docker compose file name, `TaskDefinitionLabelType`, `TaskDefinitionLabelValue` etc. And added comments for exported constants.

----

**Testing**  
- [x] Unit tests passed

**Documentation**  
- [N/A] Updated our README
----

### DIFF
--- a/ecs-cli/modules/cli/local/converter/converter.go
+++ b/ecs-cli/modules/cli/local/converter/converter.go
@@ -44,13 +44,13 @@ type LinuxParams struct {
 // See https://github.com/aws/amazon-ecs-cli/issues/797
 const SecretLabelPrefix = "ecs-local.secret"
 
+// TaskDefinitionLabelType represents the type of option used to
+// transform a task definition to a compose file e.g. remoteFile, localFile.
+// TaskDefinitionLabelValue represents the value of the option
+// e.g. file path, arn, family.
 const (
-	// taskDefinitionLabelType represents the type of option used to
-	// transform a task definition to a compose file e.g. remoteFile, localFile.
-	// taskDefinitionLabelValue represents the value of the option
-	// e.g. file path, arn, family.
-	taskDefinitionLabelType  = "ecs-local.task.type"
-	taskDefinitionLabelValue = "ecs-local.task.value"
+	TaskDefinitionLabelType  = "ecs-local.task.type"
+	TaskDefinitionLabelValue = "ecs-local.task.value"
 )
 
 // ConvertToDockerCompose creates the payload from an ECS Task Definition to be written as a docker compose file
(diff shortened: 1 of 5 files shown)


ecs-cli/modules/cli/local/converter/converter.go
```diff
@@ -44,13 +44,13 @@ type LinuxParams struct {
 // See https://github.com/aws/amazon-ecs-cli/issues/797
 const SecretLabelPrefix = "ecs-local.secret"
 
+// TaskDefinitionLabelType represents the type of option used to
```
efekarakus: Usually for constant comments we want to either:
* Give a single comment to an entire block of constants (like [here](https://github.com/golang/go/blob/master/src/text/scanner/scanner.go#L74-L88))
* Or individually give comments to each constant (like [here](https://github.com/golang/go/blob/master/src/unicode/tables.go#L5-L6))

so for you it's:
```go
const (
     // TaskDefinitionLabelType represents the type of option used to
     // transform a task definition to a compose file e.g. remoteFile, localFile.
     TaskDefinitionLabelType  = "ecs-local.task.type"

     // TaskDefinitionLabelValue represents the value of the option e.g. file path, arn, family.
     TaskDefinitionLabelValue = "ecs-local.task.value"
)
```

ecs-cli/modules/cli/local/localproject/project.go
```diff
@@ -31,11 +31,17 @@ import (
 	"github.com/urfave/cli"
 )
 
+// LocalTaskDefType represents the task is defined from a local file.
+// RemoteTaskDefType re[resents the task is defined from a remote file.
 const (
-	localTaskDefType  = "localFile"
-	remoteTaskDefType = "remoteFile"
+	LocalTaskDefType  = "localFile"
```
efekarakus: Same feedback as above here

ecs-cli/modules/cli/local/localproject/project.go
```diff
@@ -31,11 +31,17 @@ import (
...
+	RemoteTaskDefType = "remoteFile"
 )
 
+// LocalOutDefaultFileName represents the default name for the output Docker
+// Compose file.
+// LocalOutFileMode represents the file can be read/write by its owner.
+// LocalInFileName represents the default local file name for task definition JSON.
```
efekarakus: same 😂

ecs-cli/modules/cli/local/converter/converter.go
iamhopaul123: I see. Thanks for the advice. Done

PR #854, opened 2019-07-03
Write an override compose file with local create

efekarakus:
**Description of changes**: Writes a default `docker-compose.ecs-local.override.yml` if no `-o` flag is provided. Otherwise, creates a `{customName}.override.yml` file.

----

**Testing**  
- [x] Unit tests passed
- [ ] Unit tests added for new functionality

**Documentation**  
- [ ] Updated our README
----

### DIFF
--- a/ecs-cli/modules/cli/local/create_app.go
+++ b/ecs-cli/modules/cli/local/create_app.go
@@ -23,6 +23,13 @@ import (
 	"github.com/urfave/cli"
 )
 
+// TaskReadConvertWriter is the interface that groups the ReadTaskDefinition, Convert, and Write methods.
+type TaskReadConvertWriter interface {
+	ReadTaskDefinition() error
+	Convert() error
+	Write() error
+}
+
 // Create reads in an ECS task definition, converts and writes it to a local
 // Docker Compose file
 func Create(c *cli.Context) {
@@ -36,7 +43,7 @@ func Create(c *cli.Context) {
 	fmt.Printf("Successfully wrote %s\n", project.LocalOutFileName())
 }
 
-func createLocal(project localproject.LocalProject) error {
+func createLocal(project TaskReadConvertWriter) error {
 	// Reads task definition and loads it onto project
 	err := project.ReadTaskDefinition()
 	if err != nil {
--- a/ecs-cli/modules/cli/local/create_app.go
+++ b/ecs-cli/modules/cli/local/create_app.go
@@ -33,8 +33,7 @@ type TaskReadConvertWriter interface {
 func Create(c *cli.Context) {
 	project := localproject.New(c)
 
-	err := createLocal(project)
-	if err != nil {
+	if err := createLocal(project); err != nil {
 		log.Fatalf("Error with local create: %s", err.Error())
 	}
 }
--- a/ecs-cli/modules/cli/local/create_app.go
+++ b/ecs-cli/modules/cli/local/create_app.go
@@ -16,8 +16,6 @@
 package local
 
 import (
-	"fmt"
-
 	"github.com/aws/amazon-ecs-cli/ecs-cli/modules/cli/local/localproject"
 	log "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
@@ -39,8 +37,6 @@ func Create(c *cli.Context) {
 	if err != nil {
 		log.Fatalf("Error with local create: %s", err.Error())
 	}
-
-	fmt.Printf("Successfully wrote %s\n", project.LocalOutFileName())
 }
 
 func createLocal(project TaskReadConvertWriter) error {
--- a/ecs-cli/modules/cli/local/localproject/project.go
+++ b/ecs-cli/modules/cli/local/localproject/project.go
@@ -55,11 +55,10 @@ const (
 
 // LocalProject holds data needed to convert an ECS Task Definition to a Docker Compose file.
 type LocalProject struct {
-	context          *cli.Context
-	taskDefinition   *ecs.TaskDefinition
-	localBytes       []byte
-	localOutFileName string
-	inputMetadata    *converter.LocalCreateMetadata
+	context        *cli.Context
+	taskDefinition *ecs.TaskDefinition
+	localBytes     []byte
+	inputMetadata  *converter.LocalCreateMetadata
 }
 
 // New instantiates a new Local Project
@@ -74,7 +73,10 @@ func (p *LocalProject) TaskDefinition() *ecs.TaskDefinition {
 
 // LocalOutFileName returns name of compose file output by local.Create
 func (p *LocalProject) LocalOutFileName() string {
-	return p.localOutFileName
+	if customName := p.context.String(flags.Output); customName != "" {
+		return customName
+	}
+	return LocalOutDefaultFileName
 }
 
 // InputMetadata returns the metadata on the task definition used to create the docker compose file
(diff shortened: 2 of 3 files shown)


efekarakus:
# Manual tests

## Creates the default files
defaults

## Creates override file with custom name
custom-name


## New error message if no defaults are present
new-error

ecs-cli/modules/cli/local/create_app.go
```diff
@@ -16,27 +16,29 @@
...
 import (
-	"fmt"
-
 	"github.com/aws/amazon-ecs-cli/ecs-cli/modules/cli/local/localproject"
 	log "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
+// TaskReadConvertWriter is the interface that groups the ReadTaskDefinition, Convert, and Write methods.
+type TaskReadConvertWriter interface {
```
iamhopaul123: 👍 making it much more elegant

ecs-cli/modules/cli/local/localproject/project.go
```diff
@@ -200,68 +202,91 @@ var readTaskDefFromRemote = func(remote string, p *localProject) (*ecs.TaskDefin
...
 
-	_, err = out.Write(p.localBytes)
+	_, err = out.Write(content)
 
 	return err
 }
 
 // Facilitates test mocking
-var openFile func(filename string) (*os.File, error) = func(filename string) (*os.File, error) {
+var openFile = func(filename string) (*os.File, error) {
```
iamhopaul123: Just curious why do we want `openFile` to be assigned a function handle instead of using it as a function name?

ecs-cli/modules/cli/local/localproject/project.go
```diff
@@ -200,68 +202,91 @@ var readTaskDefFromRemote = func(remote string, p *localProject) (*ecs.TaskDefin
...
+	p.overrideCompose = data
 	return nil
 }
 
 // Write writes the compose data to a local compose file. The output filename is stored on the project
-func (p *localProject) Write() error {
-	// Will error if the file already exists, otherwise create
-	p.localOutFileName = LocalOutDefaultFileName
+func (p *LocalProject) Write() error {
+	if err := p.writeFile(p.LocalOutFileName(), p.baseCompose); err != nil {
```
SoManyHs: This all looks fine, but I do wonder if we are now being a little inconsistent with methods and data attached to the project struct versus passing things in as arguments (e.g. I wouldn't expect a method, i.e. `p.writeFile` to take an argument that is a member of the thing itself, i.e. `p.baseCompose`) but I could be convinced otherwise.

ecs-cli/modules/cli/local/localproject/project.go
```diff
@@ -53,48 +54,48 @@ const (
...
-	taskDefinition   *ecs.TaskDefinition
-	localBytes       []byte
-	localOutFileName string
-	inputMetadata    *converter.LocalCreateMetadata
+// LocalProject holds data needed to convert an ECS Task Definition to Docker Compose files.
+type LocalProject struct {
+	context         *cli.Context
+	taskDefinition  *ecs.TaskDefinition
+	baseCompose     []byte
+	overrideCompose []byte
```
SoManyHs: Nit: I find these two field names to be a bit ambiguous -- the original name `localBytes` was modeled after `composebytes` in [libcompose](https://github.com/aws/amazon-ecs-cli/blob/3777f934adb729b0299b53b9022eeae2ae610715/ecs-cli/vendor/github.com/docker/libcompose/project/context.go#L23). I'm wondering if it makes sense to store both the compose and override bytes together as `[][]byte`, modeling the way `readComposeFiles` works in libcompose: https://github.com/aws/amazon-ecs-cli/blob/3777f934adb729b0299b53b9022eeae2ae610715/ecs-cli/vendor/github.com/docker/libcompose/project/context.go#L23

ecs-cli/modules/cli/local/localproject/project.go
efekarakus: I believe it's so that it can be changed for unit tests https://github.com/aws/amazon-ecs-cli/blob/ecs-local/ecs-cli/modules/cli/local/localproject/project_test.go#L156-L164

ecs-cli/modules/cli/local/localproject/project.go
```diff
@@ -200,68 +202,91 @@ var readTaskDefFromRemote = func(remote string, p *localProject) (*ecs.TaskDefin
...
-	p.localOutFileName = LocalOutDefaultFileName
+func (p *LocalProject) Write() error {
+	if err := p.writeFile(p.LocalOutFileName(), p.baseCompose); err != nil {
+		return err
+	}
+	logrus.Infof("Successfully wrote %s", p.LocalOutFileName())
 
-	if fileName := p.context.String(flags.Output); fileName != "" {
-		p.localOutFileName = fileName
+	if err := p.writeFile(p.OverrideFileName(), p.overrideCompose); err != nil {
```
SoManyHs: Does this account for writing default values to the override file, or are is that coming later? Also are we always writing an override file even if one already exists?

ecs-cli/modules/cli/local/localproject/project.go
efekarakus: 🤦‍♂️ agreed, I actually didn't realize that `writeFile` and `overrideFile` were methods.
Changed them to functions so that they can be re-used for the override file.

ecs-cli/modules/cli/local/localproject/project.go
efekarakus: > Does this account for writing default values to the override file, or are is that coming later?

I'm not following 😝 do you mean if the file written has any content? If so yes, we create the content here: https://github.com/aws/amazon-ecs-cli/pull/854/files/3777f934adb729b0299b53b9022eeae2ae610715#diff-2162cb0e02fc21afb0831fe5ed0aafc5R229

> Also are we always writing an override file even if one already exists?

Yes, same behavior as the base Compose file.

ecs-cli/modules/cli/local/create_app.go
iamhopaul123: That should be a fun fact 😂 Never knew that before

ecs-cli/modules/cli/local/localproject/project.go
iamhopaul123: I see...

ecs-cli/modules/cli/local/localproject/project.go
```diff
@@ -53,48 +54,59 @@ const (
...
-	return p.localOutFileName
+func (p *LocalProject) LocalOutFileName() string {
+	if customName := p.context.String(flags.Output); customName != "" {
+		return customName
+	}
+	return LocalOutDefaultFileName
+}
+
+// OverrideFileName returns the name of the override Compose file.
+func (p *LocalProject) OverrideFileName() string {
```
SoManyHs: Can we add unit tests for this?

ecs-cli/modules/cli/local/localproject/project.go
SoManyHs: For my own edification, is there a reason to remove the type signature on the variable declaration? Is it just redundant for anonymous functions?

ecs-cli/modules/cli/local/localproject/project.go
efekarakus: Done

ecs-cli/modules/cli/local/localproject/project.go
efekarakus: I believe it's redundant if an initializer is present https://github.com/aws/amazon-ecs-cli/pull/854#discussion_r300733974

PR #915, opened 2019-08-29
feat(cicd): enable compilation through CodeBuild

hencrice:
The logic to retrieve version number and the short git hash of the last commit have been modified so that they will be retrieved during compilation. Unit tests will also run as part of the build process.

**THIS SHOULD NOT BE MERGED IN BEFORE NEXT ECS-CLI-RELEASE!**

N/A

----

**Testing**
- [N/A] Unit tests passed
- N/A] Unit tests added for new functionality
- [V] Listed manual checks and their outputs in the comments ([example]
Pushed to my own fork of this repo and the CI/CD pipeline built successfully

**Documentation**
- [N/A] Updated our README
----

### DIFF
--- a/buildspec_integ.yml
+++ b/buildspec_integ.yml
@@ -1,0 +1,7 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - make integ-test
+      - ./bin/local/ecs-cli.test


hencrice:
**THIS SHOULD NOT BE MERGED IN BEFORE THE CURRENT ECS-CLI-RELEASE!**

ecs-cli/modules/version/version.go
```diff
@@ -17,15 +17,13 @@
 // ecs-cli. It is autogenerated
 package version
 
-// Please DO NOT commit any changes to this file (specifically the hash) except
-// for those created by running ./scripts/update-version at the root of the
-// repository. Only the 'Version' const should change in checked-in source code
+// Please DO NOT commit any changes to this file.
 
-// Version is the version of the ECS CLI
-const Version = "1.15.1"
```
allisaurus: Removing this and `version-gen.go` break the `ecs-cli --version` cmd when building the binary from source:

```
// test on Windows, but same results seen using MAKE on linux

> ./bin/local/ecs-cli.exe --version
ecs-cli version  ()
```

We should retain this functionality when building outside of CodeBuild. Is there a way to move the tag/version retrieval logic from the buildspec.yml into the code somewhere so it can be reused when building locally?

Issue #933, opened 2019-09-27
ecs-cli local up cannot read from parameter store when forward slashes exist in parameter name


ndavies-om1:
#### Summary
ecs-cli local up cannot read from parameter store when forward slashes exist in parameter name

#### Description
Using command `ecs-cli local up --task-def-compose docker-compose.ecs-local.yml --override docker-compose.ecs-local.override.yml`

ecs-cli version 1.16.0

go version go1.13 darwin/amd64

macOS

#### Config files
* docker-compose.yml
```
version: "3.4"
services:
  onoff:
    command:
      - ./myapp
    environment:
      AWS_CONTAINER_CREDENTIALS_RELATIVE_URI: /creds
      AWS_REGION: us-east-1
      AWS_SDK_LOAD_CONFIG: "1"
      MY_SECRET: ${myapp_MY_SECRET}
      ECS_CONTAINER_METADATA_URI: http://169.254.170.2/v3
    image: myapp
    labels:
      ecs-local.secret.BOT_TOKEN: arn:aws:ssm:us-east-1:<myaccount>:parameter/a/b/c/my_secret
      ecs-local.task-definition-input.type: local
      ecs-local.task-definition-input.value: xxxxx/task-definition.json
    logging:
      driver: awslogs
      options:
        awslogs-group: ecs
        awslogs-region: us-east-1
        awslogs-stream-prefix: myapp
    networks:
      ecs-local-network: null
    ports:
      - 3000:3000

networks:
  ecs-local-network:
    external: true
```
* ecs-params.yml

* ~/.ecs/config

#### Observed Behavior
```
INFO[0000] The network ecs-local-network already exists 
INFO[0000] The amazon-ecs-local-container-endpoints container already exists with ID 934e9acd68cbf45d1a1c3589e5d80ce626910e5a8f2fb249e399708e7584594a 
INFO[0000] Started container with ID 934e9acd68cbf45d1a1c3589e5d80ce626910e5a8f2fb249e399708e7584594a 
FATA[0000] Failed to decrypt secret due to 
failed to retrieve decrypted secret from arn:aws:ssm:us-east-1:<myaccount>:parameter/a/b/c/my_secret due to ParameterNotFound: 
	status code: 400, request id: 1bbf17c3-8558-492f-b5a5-338b34bea5d5: ParameterNotFound: 
	status code: 400, request id: 1bbf17c3-8558-492f-b5a5-338b34bea5d5 

```

[...]

PR #935, opened 2019-09-29
Fix 933

ndavies-om1:
Fixes 933
https://github.com/aws/amazon-ecs-cli/issues/933

----

**Testing**
- [X] Unit tests passed

**Documentation**
- [N/A] Updated our README
----


ndavies-om1:
> paramName = parsedARN.Resource[len("parameter/")+1:]

That won't work as it can be passed in not as an ARN and simply as the parameter name. Tests specifically test for this, so the above change fails the unit tests

PR #938, opened 2019-10-05
Send optional fields as nil instead of "" to ECS

### LINKED ISSUE #927: errors with managed docker volume plugins with `ecs-cli compose service up`
### Summary

Running `ecs-cli compose service up` fails with a series of related errors when attempting to configure a volume that is provided by a docker managed volume plugin (like rexray/ebs and rexray/s3fs) to be used on containers in the task.

I think this is caused by incorrect documentation, bugs in code, lack-of-feature, or a combination of all. Since they are likely highly related, and the repo steps are in sequence, I will put them all in this issue.

### Setup

* Ubuntu 18.04
* ecs-cli version 1.16.0 (5c02c34)

1. Create a small EBS volume and name it `backup-laptop`. If you name is something else, then change all the below repo steps to that new name.
2. Attach and mount that EBS volume into some instance. Then write a file into that volume so you can later see that file when it is mounted again.
3. Unmount and unattach that EBS volume.
4. Create EC2 instance role, IAM task role, and IAM task execution role that is authorized for typical EC2, ECS, and EBS volume access
5. Create a cluster EC2 launch type with `ecs-cli configure`
6. Create one (1) instance in that cluster with `ecs-cli up` that uses the EC2 instance role
7. ssh to that instance
8. Install the rexray EBS volume plugin. Be aware of issue https://github.com/rexray/rexray/issues/1335
  ```
  docker plugin install --grant-all-permissions --disable rexray/ebs:0.11.4 [options...]
  docker plugin enable --timeout 120 rexray/ebs:0.11.4
  ```
9. Verify you can see your above EBS a volume with `docker volume ls`. If not, first troubleshoot that before proceeding.
10. Restart ecs agent so it will find rexray and add the capabilities
11. In all the below `ecs-params.yml` files, replace the `task_role_arn` and `task_execution_role` values with your IAM roles created in the above step

### Expected Result

In all the below Repos, the expected result is the EBS volume named `backup-laptop` to be mounted into the Alpine container at `/mnt/backup-laptop`

### Repo 1

This is the primary use case for mounting existing volumes that are managed by docker volume plugins. https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-compose-ecsparams.html writes "allows you to create docker volumes". This doesn't apply since I am not creating new volumes. The volume already exists and is available as you saw in the setup steps above.

docker-compose.yml
```yaml
version: '2'
volumes:
  backup-laptop:
    external: true
services:
  webserver:
    image: alpine
    command: tail -f /dev/null
    volumes:
      - backup-laptop:/mnt/backup-laptop
```

ecs-params.yml
```yaml
version: 1
task_definition:
  task_role_arn: ecs_task_frontend_hidale
  task_execution_role: ecs_agent_task_execution
  task_size:
    cpu_limit: 1024
    mem_limit: 500
  placement_constraints:
    - type: memberOf
      expression: attribute:ecs.capability.docker-plugin.rexray/ebs.0.11.4 exists
run_params:
  task_placement:
    constraints:
      - type: distinctInstance
```

#### Error Result for Repo 1

Fails. I acknowledge the ECS documentation isn't explicit in writing I can use externally setup docker volumes. 🤔 However, it is a primary use case for docker-compose and docker volume plugins.

```
ERRO[0000] Unable to open ECS Compose Project            error="External option is not supported"
FATA[0000] Unable to create and read ECS Compose Project  error="External option is not supported"
```

### Repo 2

Since I can't use `external` in docker-compose, I read https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-compose-ecsparams.html and it writes "The name key is required, and scope, autoprovision, driver, driver_opts and labels correspond with the Docker volume configuration fields in a task definition... Volumes defined with the docker_volumes key can be referenced in your compose file by name, even if they were not also specified in the compose file." A link leads to https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DockerVolumeConfiguration.html where it states that none of the options (autoprovision, driver, driverOpts, labels, scope) are required.

Therefore update the docker-compose and ecs-params files to not declare the `volumes:` and instead put it in the ecs-params file with `docker_volumes`.

docker-compose
```yaml
version: '2'
services:
  webserver:
    image: alpine
    command: tail -f /dev/null
    volumes:
      - backup-laptop:/mnt/backup-laptop
```

ecs-params
```yaml
version: 1
task_definition:
  task_role_arn: ecs_task_frontend_hidale
  task_execution_role: ecs_agent_task_execution
  task_size:
    cpu_limit: 1024
    mem_limit: 500
  placement_constraints:
    - type: memberOf
      expression: attribute:ecs.capability.docker-plugin.rexray/ebs.0.11.4 exists
  docker_volumes:
    - name: backup-laptop
      scope: shared
      autoprovision: false
run_params:
  task_placement:
    constraints:
      - type: distinctInstance
```

#### Error Result for Repo 2

Fails. The documentation suggests I should be able to do this. The doc writes "Volumes defined with the docker_volumes key can be referenced in your compose file by name, even if they were not also specified in the compose file". 😢 Something is errant here with both aws docs and code.

```
ERRO[0000] Unable to open ECS Compose Project            error="named volume [backup-laptop] is used but no declaration was found in the volumes section"
FATA[0000] Unable to create and read ECS Compose Project  error="named volume [backup-laptop] is used but no declaration was found in the volumes section"
dale@home-nas:/mnt/nasdata1/dale/opsworkspace/ecs-tutorial$ 
```

### Repo 3

The above error writes "named volume [backup-laptop] is used but no declaration was found in the volumes section". Therefore, I need to have the named volume in a `volumes` section yet in Repo 1 it won't allow me to declare it as `external`. So update the docker-compose to have `volumes` section with the single named volume `backup-laptop` with no subkeys.

docker-compose
```yaml
version: '2'
volumes:
  backup-laptop:
services:
  webserver:
    image: alpine
    command: tail -f /dev/null
    volumes:
      - backup-laptop:/mnt/backup-laptop
```

ecs-params
```
No change. Same file content as in Repo 2
```

#### Error Result for Repo 3

Fail. The error writes "volume.dockerVolumeConfiguration.driver should not be null or empty" yet the documentation at https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DockerVolumeConfiguration.html writes that it is optional. 🤪 This indicates that the AWS documentation is errant -- the doc should write that driver is required.

```
ERRO[0001] Error registering task definition             error="ClientException: volume.dockerVolumeConfiguration.driver should not be null or empty.\n\tstatus code: 400, request id: 423b121f-2f82-48bc-920f-22701cd099fd" family=ecs-tutorial
ERRO[0001] Create task definition failed                 error="ClientException: volume.dockerVolumeConfiguration.driver should not be null or empty.\n\tstatus code: 400, request id: 423b121f-2f82-48bc-920f-22701cd099fd"
FATA[0001] ClientException: volume.dockerVolumeConfiguration.driver should not be null or empty.
        status code: 400, request id: 423b121f-2f82-48bc-920f-22701cd099fd
```

### Repo 4

The error message writes "driver should not be null or empty". So update the ecs-params file to have the rexray driver name

docker-compose
```yaml
No change. Same file content as in Repo 3
```

ecs-params
```yaml
version: 1
task_definition:
  task_role_arn: ecs_task_frontend_hidale
  task_execution_role: ecs_agent_task_execution
  task_size:
    cpu_limit: 1024
    mem_limit: 500
  placement_constraints:
    - type: memberOf
      expression: attribute:ecs.capability.docker-plugin.rexray/ebs.0.11.4 exists
  docker_volumes:
    - name: backup-laptop
      scope: shared
      autoprovision: false
      driver: 'rexray/ebs:0.11.4'
run_params:
  task_placement:
    constraints:
      - type: distinctInstance
```

#### Success 😅 Result for Repo 4

No error appears with ecs-cli.
ssh to the instance
Get the containerId for the task's container with `docker container ls`
`docker exec -ti [containerId] sh`
`cd /mnt/backup-laptop`.
This is the EBS volume 🥳 because the file you wrote earlier into the EBS volume is there

### Notes

Related issues
https://github.com/aws/amazon-ecs-cli/issues/607
https://github.com/aws/amazon-ecs-cli/issues/699

The first related issue and the doc at https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DockerVolumeConfiguration.html suggest that I shouldn't need any entries in `ecs-params.yml` for existing volumes managed by docker volume plugins. All needed params for a task definition can be derived by `ecs-cli` and I can deal with task placement using the ecs-params `placement_constraints` you already see above.

However, I'm not adverse to ECS using the `docker_volumes.driver` for task placement. That leads me suggest the following should be supported because from it you know that name=name, scope=shared, autoprovision=false, and driver=rexray...

docker-compose
```yaml
version: '2'
volumes:
  backup-laptop:   <--- this is the name
    external: true    <--- this means autoprovision=false and scope=shared
    driver: 'rexray/ebs:0.11.4'  <--- this is the driver
  mys3bucket:   <--- this is the name
    external: true    <--- this means autoprovision=false and scope=shared
    driver: 'rexray/s3fs:0.11.4'  <--- this is the driver
services:
  webserver:
    image: alpine
    command: tail -f /dev/null
    volumes:
      - backup-laptop:/mnt/backup-laptop
```

hencrice:
Previously when sending certain requests that contain optional fields to ECS through the Go AWS SDK, we always send empty strings instead of nil for those optional fields that are not set. The service treats such an action as an explicit intent and hence fails to plug in the default values from the service side.

This PR closes #927. Specifically, the repo 3.

----

**Testing**
- [X] Unit tests passed
- [X] Integration tests passed
- [X] Listed manual checks and their outputs in the comments ([example](https://github.com/aws/amazon-ecs-cli/pull/750#issuecomment-472623042))
I was able to run `ecs-cli compose service up` instead of getting the `named volume [backup-laptop] is used but no declaration was found in the volumes section` error message described in the issue. Also verified the default, "local" driver, was correctly set on the service side in task def:
```
"volumes": [
    {
      "name": "backup-laptop",
      "host": null,
      "dockerVolumeConfiguration": {
        "autoprovision": false,
        "labels": null,
        "scope": "shared",
        "driver": "local",
        "driverOpts": null
      }
    }
  ]
```

**Documentation**
- [N/A] Updated our README
----

### DIFF
--- a/ecs-cli/modules/utils/compose/convert_task_definition.go
+++ b/ecs-cli/modules/utils/compose/convert_task_definition.go
@@ -286,8 +286,12 @@ func mergeVolumesWithoutHost(composeVolumes []string, ecsParams *ECSParams) ([]*
 		if dVol.Name != "" {
 			ecsVolume.DockerVolumeConfiguration = &ecs.DockerVolumeConfiguration{
 				Autoprovision: dVol.Autoprovision,
-				Driver:        aws.String(dVol.Driver),
-				Scope:         aws.String(dVol.Scope),
+			}
+			if dVol.Driver != nil {
+				ecsVolume.DockerVolumeConfiguration.Driver = dVol.Driver
+			}
+			if dVol.Scope != nil {
+				ecsVolume.DockerVolumeConfiguration.Scope = dVol.Scope
 			}
 			if dVol.DriverOptions != nil {
 				ecsVolume.DockerVolumeConfiguration.DriverOpts = aws.StringMap(dVol.DriverOptions)
--- a/ecs-cli/modules/utils/compose/convert_task_definition_test.go
+++ b/ecs-cli/modules/utils/compose/convert_task_definition_test.go
@@ -1383,8 +1383,8 @@ func TestConvertToTaskDefinitionWithECSParamsVolumeWithoutNameError(t *testing.T
 			DockerVolumes: []DockerVolume{
 				DockerVolume{
 					Autoprovision: aws.Bool(true),
-					Scope:         "shared",
-					Driver:        "local",
+					Scope:         aws.String("shared"),
+					Driver:        nil,
 					DriverOptions: options,
 					Labels:        labels,
 				},
--- a/ecs-cli/modules/utils/compose/ecs_params_reader.go
+++ b/ecs-cli/modules/utils/compose/ecs_params_reader.go
@@ -75,9 +75,9 @@ type ContainerDef struct {
 
 type DockerVolume struct {
 	Name          string            `yaml:"name"`
-	Scope         string            `yaml:"scope"`
+	Scope         *string           `yaml:"scope"`
 	Autoprovision *bool             `yaml:"autoprovision"`
-	Driver        string            `yaml:"driver"`
+	Driver        *string           `yaml:"driver"`
 	DriverOptions map[string]string `yaml:"driver_opts"`
 	Labels        map[string]string `yaml:"labels"`
 }
--- a/ecs-cli/modules/utils/compose/ecs_params_reader_test.go
+++ b/ecs-cli/modules/utils/compose/ecs_params_reader_test.go
@@ -647,9 +647,9 @@ task_definition:
 	expectedVolumes := []DockerVolume{
 		DockerVolume{
 			Name:          "my_volume",
-			Scope:         "shared",
+			Scope:         aws.String("shared"),
 			Autoprovision: aws.Bool(true),
-			Driver:        "doggyromcom",
+			Driver:        aws.String("doggyromcom"),
 			DriverOptions: map[string]string{
 				"pudding": "is-engaged-to-marry-Tum-Tum",
 				"clyde":   "professes-his-love-at-the-ceremony",


sonofachamp:
Can we use `omitempty` [yaml struct tags](https://godoc.org/gopkg.in/yaml.v2) on the fields?

> omitempty - Only include the field if it's not set to the zero
             value for the type or to empty slices or maps.
             Zero valued structs will be omitted if all their public
             fields are zero, unless they implement an IsZero
             method (see the IsZeroer interface type), in which
             case the field will be included if that method returns true.

kohidave:
Is there any way folks may have come to depend on this empty string behavior? I guess what I'm asking is if we launch this, are a bunch of folks going to see resources being created or destroyed and be surprised?

hencrice:
> Can we use `omitempty` [yaml struct tags](https://godoc.org/gopkg.in/yaml.v2) on the fields?
> 
> > omitempty - Only include the field if it's not set to the zero
> > value for the type or to empty slices or maps.
> > Zero valued structs will be omitted if all their public
> > fields are zero, unless they implement an IsZero
> > method (see the IsZeroer interface type), in which
> > case the field will be included if that method returns true.

I'm afraid `omitempty` is only applicable during the `Marshal` function provided by the `yaml` package. Since we are simply doing in-memory transformation, adding that tag does not seem to help.

hencrice:
> Is there any way folks may have come to depend on this empty string behavior? I guess what I'm asking is if we launch this, are a bunch of folks going to see resources being created or destroyed and be surprised?

1. For those customers who already provide non-empty strings for these fields, they will be ok.
2. For the others who have been providing empty strings, they would not have successfully registered their task definitions (same as repo3 in #927 ). I'm leaning toward fixing this because I think it's a bug that prevents the legitimate use case which relies on the defaults provided by the service. This change will only impact those customers who explicitly rely on the 400 user error and in that worst case, resources will be created instead of destroyed.

PR #970, opened 2019-12-31
fixing local create to use taskRole

### LINKED ISSUE #969: ecs-cli local create - taskRoleArn not used by local container
When creating docker-compose files using ecs-cli local create, the taskRole is not used by the local container.

#### Summary
`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is always set to `creds` regardless of whether a taskRole is used or not. When I want to run a container locally, it should use the taskRole defined by default. If I want to override it to use a different set of creds, then I should do that in the `docker-compose.ecs-local.override.yml`.

ndavies-om1:
Looks for a TaskRole and sets that as the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environment variable. Also deletes the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` from the `docker-compose.ecs-local.override.yml` as this is not necessary and will always override this setting

Fixes #969

----

**Testing**
- [X] Unit tests passed
- [ ] Integration tests passed
- [ ] Listed manual checks and their outputs in the comments ([example](https://github.com/aws/amazon-ecs-cli/pull/750#issuecomment-472623042))

**Documentation**
- [ ] Updated our README
----

### DIFF
--- a/ecs-cli/modules/cli/local/converter/converter.go
+++ b/ecs-cli/modules/cli/local/converter/converter.go
@@ -26,7 +26,7 @@ import (
 	composeV3 "github.com/docker/cli/cli/compose/types"
 	"github.com/docker/go-units"
 	log "github.com/sirupsen/logrus"
-	arnParser "github.com/aws/aws-sdk-go/aws/arn"
+	arnparser "github.com/aws/aws-sdk-go/aws/arn"
 )
 
 // LinuxParams is a shim between members of ecs.LinuxParamters and their
@@ -410,8 +410,8 @@ func convertEnvironment(def *ecs.ContainerDefinition, taskRoleARN string) map[st
 	}
 
     credsName := endpointsTempCredsPath
-    if parsedRoleARN, err := arnParser.Parse(taskRoleARN); taskRoleARN != "" && err == nil {
-        credsName = "/" + parsedRoleARN.Resource
+    if parsedRoleARN, err := arnparser.Parse(taskRoleARN); taskRoleARN != "" && err == nil {
+        credsName = "/" + parsedRoleARN.Resource // The parsed resource is formatted as "role/{roleName}"
     }
 
 	out[ecsCredsProviderEnvName] = aws.String(credsName)
--- a/ecs-cli/modules/cli/local/converter/converter.go
+++ b/ecs-cli/modules/cli/local/converter/converter.go
@@ -26,6 +26,7 @@ import (
 	composeV3 "github.com/docker/cli/cli/compose/types"
 	"github.com/docker/go-units"
 	log "github.com/sirupsen/logrus"
+	arnParser "github.com/aws/aws-sdk-go/aws/arn"
 )
 
 // LinuxParams is a shim between members of ecs.LinuxParamters and their
@@ -49,9 +50,10 @@ type LocalCreateMetadata struct {
 // CommonContainerValues contains values for top-level Task Definition fields
 // that apply to all containers within the task definition
 type CommonContainerValues struct {
-	Ipc     string
-	Pid     string
-	Volumes []*ecs.Volume
+	Ipc         string
+	Pid         string
+	Volumes     []*ecs.Volume
+	TaskRoleArn string
 }
 
 const (
@@ -143,9 +145,10 @@ func createComposeServices(taskDefinition *ecs.TaskDefinition, metadata *LocalCr
 	}
 
 	commonValues := &CommonContainerValues{
-		Pid:     pid,
-		Ipc:     ipc,
-		Volumes: taskDefinition.Volumes,
+		Pid:         pid,
+		Ipc:         ipc,
+		Volumes:     taskDefinition.Volumes,
+		TaskRoleArn: aws.StringValue(taskDefinition.TaskRoleArn),
 	}
 
 	if len(taskDefinition.ContainerDefinitions) < 1 {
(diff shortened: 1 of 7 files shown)


ecs-cli/modules/cli/local/converter/converter.go
```diff
@@ -406,7 +409,12 @@ func convertEnvironment(def *ecs.ContainerDefinition) map[string]*string {
 		out[secretName] = &shellEnv
 	}
 
-	out[ecsCredsProviderEnvName] = aws.String(endpointsTempCredsPath)
+    credsName := endpointsTempCredsPath
+    if parsedRoleARN, err := arnParser.Parse(taskRoleARN); taskRoleARN != "" && err == nil {
+        credsName = "/" + parsedRoleARN.Resource
```
efekarakus: oh I see `parsedRoleARN.Resource` is already prefixed with `"role/"` and that's why that works, is that correct?

What do you think of adding a comment here like: `//  The parsed resource is formatted as "role/{roleName}"`

efekarakus: On second thought, I believe this was a conscious decision on our end to always set it to `"/creds"` because of this blurb:
> Note: We do not recommend using production credentials or production roles when testing locally. Modifying the trust policy of a production role changes its security boundary. More importantly, using credentials with access to production when testing locally could lead to accidental changes in your production account. We recommend using a separate account for testing.

https://github.com/awslabs/amazon-ecs-local-container-endpoints/blob/master/docs/features.md#vend-credentials-to-containers

I think to allow this use-case and maintain the security bar, I'd be okay with adding a new flag to `local create` for the user to acknowledge of the risk they're taking. Maybe something like: `local create -t myTaskDef:1 --role`
```
$ ecs-cli local create --help
...
--role  Uses the task role ARN instead of temporary credentials.
```

What do you think?

ndavies-om1: I understand where you are coming from regarding the production credentials, but it assumes that the local user can assume the production credentials (the task role), which they should not be able to do as they are production credentials. 

When running the ecs-cli local up command, the `amazon-ecs-local-container-endpoints` container uses the user's `default` profile, which typically is the users AWS account. So for the app container to use the task role from the task definition, their user account would need to be able to assume the task role, which is unlikely unless the users account has a particularly bad security setup, such running as root or user having full admin rights AND their production account is the same account as the users account. 

To use a role that had assume privileges to the task role, they'd have to set the AWS_PROFILE environment variable for the `amazon-ecs-local-container-endpoints` container to use a profile that could assume the task role to create the creds used by the app container.

I feel that getting an accurate docker-compose file that reflects what is needed to run a container locally mimicking what is running in AWS as closely as possible is what a user wants from this command. 

The options are follows:
- Keeping the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI: /creds` setting in the override file. This is what will get used by the user when they run locally and it will meet the not using production creds requirement. They can then explicitly delete the override file if they want to run it with the task role
- Use the `--task` option as suggested. However the user needs to go and lookup what the tasks role is before pulling it locally.
- Keep the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI: /creds`, but also add the task role to the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` but commented out so it is clear what needs to change to use the task role. This would then look like
```
environment:
      AWS_CONTAINER_CREDENTIALS_RELATIVE_URI: /creds # to use the task role replace this with "/role/taskrole"
```
- Keep the PR as is, and acknowledge that if the user has a really disastrous IAM/security setup they could run as production locally without realizing. 

Thoughts?

allisaurus: > Use the --task option as suggested. However the user needs to go and lookup what the tasks role is before pulling it locally.

My understanding of @efekarakus 's suggestion is that this would be a boolean flag to indicate whether or not to use the task role present in the task definition, in which case looking up the role wouldn't be necessary (to use, anyway - as a user, always wise to make sure you understand what it will be before enacting it).

If that's indeed the case, I would be inclined to go with option 2 as it seems to have the desired effect without requiring users to manually edit the files (though I would suggest a more descriptive flag name, e.g., `--use-role`)

efekarakus: @allisaurus Correct! That's what I meant. The name of my flag was poor, `--use-role` is more sensible for a boolean flag :)

So the code proposed can stay since it removes the burden of looking up the task role, but it'd only be enabled if the user passes a `--use-role` flag to acknowledge the potential risk they're taking.

ndavies-om1: I pushed up changes to make it work as per option 2. Tests updated too

ecs-cli/modules/cli/local/converter/converter.go
```diff
@@ -175,6 +179,14 @@ func createComposeServices(taskDefinition *ecs.TaskDefinition, metadata *LocalCr
 	return services, nil
 }
 
+func resolveCredentials(taskRoleARN string, useRole bool) string {
+    credsName := endpointsTempCredsPath
+    if parsedRoleARN, err := arnparser.Parse(taskRoleARN); useRole && taskRoleARN != "" && err == nil {
+        credsName = "/" + parsedRoleARN.Resource // The parsed resource is formatted as "role/{roleName}"
+    }
+    return credsName
```
efekarakus: We can leverage [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) to make the conditional easier to parse:
```go
if !useRole {
  return endpointsTempCredsPath
}
parsedARN, err := arnparser.Parse(taskRoleARN)
if err != nil {
  return endpointsTempCredsPath
}
return "/" + parsedARN.Resource
```

PR #980, opened 2020-01-22
update docker/cli dependency constraint

### LINKED ISSUE #978: dep Gopkg.toml Gopkg.lock, and vendor dependencies not in tight sync
Files checked into this repo in the `vendor` directory are not in tight sync with `Gopkg.toml`. In addition, `Gopkg.toml` and `Gopkg.lock` may not be in sync with each other. (Caveat, I am newish to go and dep...and SemVer ranges are involved...so perhaps there is something I have misunderstood.)

### Description

To start, two examples that expose the out-of-sync, or my lack of understanding 😉

#### `Gopkg.toml` and `Gopkg.lock` out of sync with each other

Reference the `version` value in

https://github.com/aws/amazon-ecs-cli/blob/d1685dd9bbd4e3d9375a7e076f69cc20ec64c13b/ecs-cli/Gopkg.toml#L79-L81

and now in

https://github.com/aws/amazon-ecs-cli/blob/d1685dd9bbd4e3d9375a7e076f69cc20ec64c13b/ecs-cli/Gopkg.lock#L126-L139

They do not match. However, due to SemVer (see Notes below) this may be technically possible. However, is this desired? Does the team actually want any (version >18.6.0) but not the 19.x.x updates? 🤔 Or perhaps someone didn't realize how dep's resolver would work here.

#### `vendor/` files out of sync with `Gopkg.*`

Next, look at the contents of the following file from that `github.com/docker/cli` dependency. `Gopkg.toml` declares that this dependency should be tag `18.06.0-ce` or greater in the 18.x.x range. Here is a specific line of a file in that dependency at that tag

https://github.com/docker/cli/blob/2cb2e84287c916d484cf5d6fc93987ba537f3aab/cli/compose/loader/loader.go#L268

```go
func transform(source map[string]interface{}, target interface{}) error {
```

Now compare that line to the file currently checked into the `vendor/` directory

https://github.com/aws/amazon-ecs-cli/blob/d1685dd9bbd4e3d9375a7e076f69cc20ec64c13b/ecs-cli/vendor/github.com/docker/cli/cli/compose/loader/loader.go#L270

As you can see, these are different files. The file checked into `vendor` does not match the dependency declaration. Instead, that file is something newer than tag `18.06.0-ce`.

### Notes

I see activity by two people on or arounds these dep file entries and this specific dependency. They may have some knowledge: @efekarakus and @allisaurus

Perhaps contributing to this, is how constraints have been declared in `Gopkg.toml`. There are several that are not frozen at a specific version and instead will update to whatever is the latest version using dep and SemVer rules. For example, the dependency above is declared

https://github.com/aws/amazon-ecs-cli/blob/d1685dd9bbd4e3d9375a7e076f69cc20ec64c13b/ecs-cli/Gopkg.toml#L79-L81

Because of the dep and SemVer rules, `version = "18.06.0-ce"` is made into a canonical constraint `^18.6.0-ce`. Correct. It drops the `0` and prepends the `^`. The caret pins the left-most non-zero digit in the constraint. So in this case, it means: (version >= 18.6.0) and (version < 19.0.0). I tested and that constraint on that dependency installed `docker/cli` tag `v18.09.9`. 🤹‍♀️

If it is desired to have a specific tag and nothing newer, then the constraint should be either `version = "=18.06.0-ce"` or `revision = "v18.06.0-ce"`

Using this method and some hunting, I determined that the files checked into `vendor/` are **not** from the newest possible with the current constraint. Instead, they are likely from tag `v18.09.5`. That suggests that someone manually copied files (hope not) or ran `dep ensure -update github.com/docker/cli` in the period of time that `v18.09.5` was the newest. If this command was run today, it would pull `v18.09.9` which has different files.

diablodale:
Fixes #978 to more clearly specify the `docker/cli` dependency constraint
This change specifies to use any patch from `18.09.x` and includes the `Gopkg.*` and `vendor/` files to match.
My thoughts on this below. 🧐

### Investigation

The Docker projects often do **not** use SemVer. Instead, they use a `YY.MM.patch` approach as documented https://www.docker.com/blog/docker-enterprise-edition/. This is true for this `docker/cli` dependency and others like `docker/docker`

Refer to issue #978 for diagnostic details. In addition to those diagnostics...

The `vendor/` files for this `docker/cli` dependency that were commited months ago to this repo are from `18.09.5`. This was verified with two methods: 1) `Gopkg.lock` declares the specific version and commit it locked; 2) `git diff` showed no changed files when I updated the `Gopkg.toml` constraint from `18.06.0-ce` to `=18.09.5` and ran `dep ensure -update github.com/docker/cli`

### Questions

I have a cautious query, in the `Gopkg.toml` is an override for the `docker/docker` dependency and it is specifically pinned to a commit that is not any full release. Instead, it appears to be a intermediate commit chain before the `18.06.0` release. While this obviously doesn't cause problems today, it is suspicious to use a pre-18.06.0 codebase with elsewhere post-18.06.0 code.

### Recommendations

To keep within a grouping of engine/api compatibility, I recommend to lock on a specific `YY.MM` like `18.09` and then allow the most recent patch. This follows somewhat the approach of @PettitWesley when he wrote "personally prefer to use the latest version of dependencies, unless backwards breaking changes force us to stick with an older version".

I recommend that you do allow patch/bugfixes. Therefore instead of using a `^` constraint, I recommend using a `~` constraint at the patch level. We already know `18.09.5` works because `ecs-cli` has been using/released with that for many months. There have been some minor patches to the `18.09` release and it is currently up to `18.09.9`. You can see those changes in this PR as instructed by CONTRIBUTING.md.

It is also probably desired to support the release e.g. `18.09` that the ECS infrastructure also supports. Currently, the ECS optimized AMIs use docker `18.09`

This also prepares the `ecs-cli` project for a forthcoming patch to the `docker/cli` dependency which I already have in queue on that project. That patch will correct one function that isn't following the compose schema correctly and will enable a new ecs-cli feature that I've completed coding. That work is documented in 607 which I don't want github to autolink that issue.

----

### Checklist

**Testing**
- [X] Unit tests passed
- [X] Listed manual checks and their outputs in the comments

**Documentation**
- [N/A] Updated our README
----

### Comments

I `make docker-build` with no errors. It generated an `ecs-cli` which successfully ran some random commands I tested.

### DIFF
--- a/ecs-cli/Gopkg.lock
+++ b/ecs-cli/Gopkg.lock
@@ -124,7 +124,7 @@
   version = "v1.1.0"
 
 [[projects]]
-  digest = "1:a5ca15953ec7d39eced701d326903a1ffed0a7f1339a9b02156f6bc6b2a7cefd"
+  digest = "1:ae25b86119af299d4f1c95a91180580e3ab5a4fd0293ec423dac1c304050a944"
   name = "github.com/docker/cli"
   packages = [
     "cli/compose/interpolation",
@@ -135,8 +135,8 @@
     "opts",
   ]
   pruneopts = "UT"
-  revision = "c89750f836c57ce10386e71669e1b08a54c3caeb"
-  version = "v18.09.5"
+  revision = "1752eb3626e3a17b0881135a9402e57728208fed"
+  version = "v18.09.9"
 
 [[projects]]
   digest = "1:4ddc17aeaa82cb18c5f0a25d7c253a10682f518f4b2558a82869506eec223d76"
--- a/ecs-cli/Gopkg.toml
+++ b/ecs-cli/Gopkg.toml
@@ -78,7 +78,7 @@
 
 [[constraint]]
   name = "github.com/docker/cli"
-  version = "18.06.0-ce"
+  version = "~18.09.5"
 
 [[override]]
   name = "github.com/docker/docker"
--- a/ecs-cli/vendor/github.com/docker/cli/cli/compose/loader/interpolate.go
+++ b/ecs-cli/vendor/github.com/docker/cli/cli/compose/loader/interpolate.go
@@ -16,6 +16,8 @@ var interpolateTypeCastMapping = map[interp.Path]interp.Cast{
 	servicePath("deploy", "replicas"):                                toInt,
 	servicePath("deploy", "update_config", "parallelism"):            toInt,
 	servicePath("deploy", "update_config", "max_failure_ratio"):      toFloat,
+	servicePath("deploy", "rollback_config", "parallelism"):          toInt,
+	servicePath("deploy", "rollback_config", "max_failure_ratio"):    toFloat,
 	servicePath("deploy", "restart_policy", "max_attempts"):          toInt,
 	servicePath("ports", interp.PathMatchList, "target"):             toInt,
 	servicePath("ports", interp.PathMatchList, "published"):          toInt,
--- a/ecs-cli/vendor/github.com/docker/cli/cli/compose/loader/loader.go
+++ b/ecs-cli/vendor/github.com/docker/cli/cli/compose/loader/loader.go
@@ -476,12 +476,13 @@ func resolveVolumePaths(volumes []types.ServiceVolumeConfig, workingDir string, 
 		}
 
 		filePath := expandUser(volume.Source, lookupEnv)
-		// Check for a Unix absolute path first, to handle a Windows client
-		// with a Unix daemon. This handles a Windows client connecting to a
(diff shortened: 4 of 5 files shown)


PettitWesley:
This LGTM. Can someone else ship it/review so we can merge?

cc @allisaurus, since you looked at the issue as well

allisaurus:
re: `docker/docker` override, we should reevaluate whether this is still needed or at least up the revision to something closer to latest. To help isolate your change from the potential fallout of that, I've created #981 to track this.

Otherwise: LGTM! Thanks a bunch for working on this.

PR #1000, opened 2020-03-13
fix(pipeline): Add verification step and integ tests to pipeline

bvtujo:
Adds integ test stage in parallel with build stage which runs "make integ"
Adds verification stage after release which uses the following logic: 
* Download artifacts from cn-north-1 endpoint 
* Download signatures from US endpoint
* If signature mismatch, or unable to download artifact, retry in 15 minutes
* Repeat until 8 hour codebuild project timeout.  

----

**Testing**
- [x ] Unit tests passed
- [x ] Integration tests passed

----

### DIFF
--- a/buildspec_canary.yml
+++ b/buildspec_canary.yml
@@ -6,7 +6,6 @@ phases:
       golang: 1.13
   build:
     commands:
-      - GOPROXY=direct
       - latestTag=$(git describe --tags `git rev-list --tags --max-count=1`)
       - echo "checking out $latestTag"
       - git checkout $latestTag
--- a/buildspec_canary.yml
+++ b/buildspec_canary.yml
@@ -1,0 +1,14 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      golang: 1.13
+  build:
+    commands:
+      - GOPROXY=direct
+      - latestTag=$(git describe --tags `git rev-list --tags --max-count=1`)
+      - echo "checking out $latestTag"
+      - git checkout $latestTag
+      - make integ-test
+      - ./bin/local/ecs-cli.test
--- a/buildspec_integ.yml
+++ b/buildspec_integ.yml
@@ -12,7 +12,9 @@ phases:
       - cd /go/src/github.com/aws/amazon-ecs-cli/
   build:
     commands:
-      - $TMPDIR=/tmp make integ-test
+      - mkdir ./test_data
+      - export TMPDIR=./test_data 
+      - make integ-test
       - ./bin/local/ecs-cli.test
 artifacts:
   files:
--- a/buildspec_integ.yml
+++ b/buildspec_integ.yml
@@ -11,8 +11,6 @@ phases:
       - cd /go/src/github.com/aws/amazon-ecs-cli/
   build:
     commands:
-      - mkdir tmp
-      - export TMPDIR=tmp
       - make integ-test
       - ./bin/local/ecs-cli.test
 artifacts:
--- a/buildspec_integ.yml
+++ b/buildspec_integ.yml
@@ -4,11 +4,12 @@ phases:
   install:
     runtime-versions:
       golang: 1.13
-  build:
+  pre_build:
     commands:
       - GOPROXY=direct
-      - latestTag=$(git describe --tags `git rev-list --tags --max-count=1`)
-      - echo "checking out $latestTag"
-      - git checkout $latestTag
+      - mkdir -p /go/src/github.com/aws/amazon-ecs-cli
+      - cp -R . /go/src/github.com/aws/amazon-ecs-cli/
+      - cd /go/src/github.com/aws/amazon-ecs-cli/
+  build:
+    commands:
       - make integ-test
-      - ./bin/local/ecs-cli.test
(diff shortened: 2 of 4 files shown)


buildspec_canary.yml
```diff
@@ -0,0 +1,14 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      golang: 1.13
+  build:
+    commands:
+      - GOPROXY=direct
```
hencrice: nit: can this be part of env vars? Also, I think this is only needed on Amazon internal network. Should technically able to call Go package registry on Codebuild.

buildspec_canary.yml
```diff
@@ -0,0 +1,14 @@
+version: 0.2
```
hencrice: Don't forget to change https://tiny.amazon.com/ne2o910k/IsenLink to use this new buildspec for canary

buildspec_integ.yml
```diff
@@ -1,10 +1,15 @@
...
+      - cd /go/src/github.com/aws/amazon-ecs-cli/
   build:
     commands:
-      - latestTag=$(git describe --tags `git rev-list --tags --max-count=1`)
-      - echo "checking out $latestTag"
-      - git checkout $latestTag
       - make integ-test
-      - ./bin/local/ecs-cli.test
\ No newline at end of file
+      - ./bin/local/ecs-cli.test
```
hencrice: How does this buildspec know `./bin/local/ecs-cli.test` must exist? Also, how & where is the test binary built in the pipeline?

buildspec_integ.yml
```diff
@@ -1,10 +1,15 @@
...
 
 phases:
+  install:
+    runtime-versions:
+      golang: 1.13
+  pre_build:
+    commands:
+      - mkdir -p /go/src/github.com/aws/amazon-ecs-cli
+      - cp -R . /go/src/github.com/aws/amazon-ecs-cli/
+      - cd /go/src/github.com/aws/amazon-ecs-cli/
```
hencrice: why do we need to make these directories when we are just going to run the test binary below?

buildspec_canary.yml
bvtujo: Yes, will change this in the CM step to update the canary's env vars.

buildspec_canary.yml
bvtujo: this is a CM step!

buildspec_integ.yml
bvtujo: I fought with codebuild for quite a long time, and this was the only way I could get the make target to run properly--otherwise I was getting errors where go could not find the dependencies, could not find the test coverage directories for the call to gocovmerge, was failing due to type mismatches stemming from the go dependencies being in a different directory...

buildspec_canary.yml
bvtujo: Removed in most recent commit.

buildspec_integ.yml
bvtujo: I'm not sure how we know that about the canary, to be honest, but this succeeds in codebuild


```
[Container] 2020/03/13 03:57:20 Running command ./bin/local/ecs-cli.test
853 | NAME:
854 | ecs-cli - Command line interface for Amazon ECS
855 |  
856 | USAGE:
857 | ecs-cli.test [global options] command [command options] [arguments...]
858 |  
859 | VERSION:
860 | 1.18.1 (*UNKNOWN)
861 |  
862 | AUTHOR:
863 | Amazon Web Services
864 |  
865 | COMMANDS:
866 | configure         Stores a single cluster configuration.
867 | up                Creates the ECS cluster (if it does not already exist) and the AWS resources required to set up the cluster.
868 | down              Deletes the CloudFormation stack that was created by ecs-cli up and the associated resources.
869 | scale             Modifies the number of container instances in your cluster. This command changes the desired and maximum instance count in the Auto Scaling group created by the ecs-cli up command. You can use this command to scale up (increase the number of instances) or scale down (decrease the number of instances) your cluster.
870 | ps                Lists all of the running containers in your ECS cluster.
871 | push              Pushes an image to an Amazon ECR repository.
872 | pull              Pulls an image from an Amazon ECR repository.
873 | images            Lists images from an Amazon ECR repository. Lists all images in all repositories by default.
874 | license           Prints the LICENSE files for the ECS CLI and its dependencies.
875 | compose           Executes docker-compose-style commands on an ECS cluster.
876 | check-attributes  Checks if a given list of container instances can run a given task definition by checking their attributes. Outputs attributes that are required by the task definition but not present on the container instances.
877 | logs              Retrieves container logs from CloudWatch logs. Assumes your Task Definition uses the awslogs driver and has a log stream prefix specified.
878 | registry-creds    Facilitates the creation and use of private registry credentials within ECS.
879 | local             Runs your ECS tasks locally.
880 | help, h           Shows a list of commands or help for one command
881 |  
882 | GLOBAL OPTIONS:
883 | --endpoint value  Use a custom endpoint with the ECS CLI
884 | --help, -h        show help
885 | --version, -v     print the version
886 | PASS
887 | coverage: 2.1% of statements in ./ecs-cli/modules/...
888

```

buildspec_integ.yml
hencrice: lol ok

buildspec_integ.yml
hencrice: Missed the `make integ-test` statement on the previous line. Please ignore

PR #1013, opened 2020-04-16
Feature for attaching multiple target groups in a service

### LINKED ISSUE #995: Add Support for Multiple Target Group
#### Context
Currently as noted [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-compose-service-up.html), `--target-group-arn` is a string which restricts attaching more target groups to an ECS service.

To solve this issue and support attaching multiple target groups, I think we can make another wrapped up field called `--load-balancers`, which is a list of struct consisting of `--container-name`, `--container-port`, and `--target-group-arn` so that we can attach the service to those target groups.

#### Alternatives

Do the attachment by using console or calling APIs.

#### Has the feature been requested before?

Related to https://github.com/aws/amazon-ecs-cli/issues/910.

seongm-1:
The ability to register multiple target groups in a single service

Resolves #995 

----

**Testing**
- [x] Unit tests passed
- [ ] Integration tests passed
- [ ] Listed manual checks and their outputs in the comments ([example](https://github.com/aws/amazon-ecs-cli/pull/750#issuecomment-472623042))

**Documentation**
- [ ] Updated our README
----


ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -121,6 +123,9 @@ func (s *Service) LoadContext() error {
 		if targetGroupArn != "" && loadBalancerName != "" {
 			return errors.Errorf("[--%s] and [--%s] flags cannot both be specified", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
 		}
+		if len(loadBalancerList) != 0 {
```
iamhopaul123, line 126: You need to specify role flag to allow Amazon ECS to make calls to your load balancer or target group on your behalf. It seems like if you do the validation like this users cannot do that :P

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -133,6 +138,16 @@ func (s *Service) LoadContext() error {
 			s.loadBalancer.LoadBalancerName = aws.String(loadBalancerName)
 		}
 		s.role = role
+		s.loadBalancers = append(s.loadBalancers, s.loadBalancer)
+	} else if len(loadBalancerList) != 0 {
+		if len(loadBalancerList) > 5 {
```
iamhopaul123, line 143: oh we don't do the validation for this, because this might change and differs from account to account.

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -475,6 +490,19 @@ func (s *Service) GetTags() ([]*ecs.Tag, error) {
 	return s.tags, nil
 }
 
+func (s *Service) getLoadBalancerList(loadBalancerValue []string) ([]*ecs.LoadBalancer, error) {
+	if s.loadBalancers == nil {
+		loadBalancers := make([]*ecs.LoadBalancer, 0)
```
iamhopaul123, line 495: Usually in golang we prefer to init an empty array like this

```
var loadBalancers []*ecs.LoadBalancer
```

unless you know the length of the slice you can do

```
loadBalancers := make([]*ecs.LoadBalancer, len(loadBalancerValue)
```

Refer to https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -475,6 +490,19 @@ func (s *Service) GetTags() ([]*ecs.Tag, error) {
...
+		var err error
+		loadBalancers, err = utils.ParseLoadBalancers(loadBalancerValue, loadBalancers)
```
iamhopaul123, line 497: you don't actually need to init an empty slice. Instead it would be cleaner to

```
loadBalancers, err := utils.ParseLoadBalancers(loadBalancerValue)
```

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -475,6 +490,19 @@ func (s *Service) GetTags() ([]*ecs.Tag, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		s.loadBalancers = loadBalancers
```
iamhopaul123, line 501: This is confusing, since you also assign value to `s.loadBlancers` outside of this function. It ends up being `s.LoadBalancers = s.LoadBalancers`.
 
I think you probably don't need `getLoadBalancerList` and let `utils.ParseLoadBalancers` takes the slice of string as input and slice of struct as output.

ecs-cli/modules/commands/compose/service/compose_service_command.go
```diff
@@ -226,6 +226,7 @@ func loadBalancerFlags() []cli.Flag {
 	containerNameUsageString := fmt.Sprintf("[Optional] Specifies the container name (as it appears in a container definition). This parameter is required if --%s or --%s is specified.", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
```
iamhopaul123, line 226: Don't forget [deprecated] in the usage for old flags :)

ecs-cli/modules/commands/compose/service/compose_service_command.go
```diff
@@ -226,6 +226,7 @@ func loadBalancerFlags() []cli.Flag {
...
 	containerPortUsageString := fmt.Sprintf("[Optional] Specifies the port on the container to associate with the load balancer. This port must correspond to a containerPort in the service's task definition. This parameter is required if --%s or --%s is specified.", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
 	loadBalancerNameUsageString := fmt.Sprintf("[Optional] Specifies the name of a previously configured Classic Elastic Load Balancing load balancer to associate with your service. NOTE: For Application Load Balancers or Network Load Balancers, use the --%s flag.", flags.TargetGroupArnFlag)
+	loadBalancerListUsageString := fmt.Sprintf("[Optional] Specifies multiple load balancers to register with a service. Can not be used with --%s flag or --%s at the same time. ", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
```
iamhopaul123, line 229: nit: "Cannot"

ecs-cli/modules/commands/flags/flags.go
```diff
@@ -134,6 +134,7 @@ const (
 	RoleFlag                                = "role"
 	ComposeServiceTimeOutFlag               = "timeout"
 	ForceDeploymentFlag                     = "force-deployment"
+	LoadBalancerListFlag                    = "load-balancer-list"
```
iamhopaul123, line 137: What do you think of make the flag name to be "target-groups", since this flag is to allow users to register their service to multiple target groups.

ecs-cli/modules/utils/utils.go
```diff
@@ -79,7 +80,7 @@ func ParseTags(flagValue string, tags []*ecs.Tag) ([]*ecs.Tag, error) {
 	return tags, nil
 }
 
-// GetTags parses AWS Resource tags from the flag value
+// GetTagsMap parses AWS Resource tags from the flag value
```
iamhopaul123, line 83: Nice!

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,40 @@ func GetPartition(region string) string {
 		return "aws"
 	}
 }
+
+// ParseLoadBalancers returns a list of load balacners struct
```
iamhopaul123, line 111: nit: typo for "balancers", and usually in the comment we just specify the usage for this function: only describe what this function does. Also can include an input example here like ["targetGroupArn=....,containerName=....", "...."] so that the others can be aware of the logic below.

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,40 @@ func GetPartition(region string) string {
...
+// For users wanting to register multiple or one load balancers to a single service
+func ParseLoadBalancers(flagValue []string, list []*ecs.LoadBalancer) ([]*ecs.LoadBalancer, error) {
+
```
iamhopaul123, line 114: Same as comment above, you can actually declare an empty `[]*ecs.LoadBalancer` here then return this variable.

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,40 @@ func GetPartition(region string) string {
...
+	for i := 0; i < len(flagValue); i++ {
```
iamhopaul123, line 115: Would be better to be like this

```
for _, flagValue := range flagValues {
  ...
}
```

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,40 @@ func GetPartition(region string) string {
...
+		m := make(map[string]string)
+		targetGroupArn := false
```
iamhopaul123, line 117: maybe "elbv2" as var name because this is supposed to determine if this is registering to classic LB or ALB/NLB

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,40 @@ func GetPartition(region string) string {
...
+		keyValPairs := strings.Split(flagValue[i], ",")
+
+		for _, kv := range keyValPairs {
+			pair := strings.SplitN(kv, "=", 2)
```
iamhopaul123, line 121: Needs a validation here. What if input is "targetGroupArn=asdf=asdf".

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,40 @@ func GetPartition(region string) string {
...
+			m[pair[0]] = pair[1]
+			if pair[0] == "targetGroupArn" {
+				targetGroupArn = true
+			}
+		}
+		containerPort, err := strconv.ParseInt(m["containerPort"], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("Please pass integer value for the flag %s", m["containerPort"])
```
iamhopaul123, line 129:
```suggestion
			return nil, fmt.Errorf("Fail to parse container port %s for container %s ", m["containerPort"], m["containerName"])
```

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,40 @@ func GetPartition(region string) string {
...
+		keyValPairs := strings.Split(flagValue[i], ",")
+
+		for _, kv := range keyValPairs {
+			pair := strings.SplitN(kv, "=", 2)
+			m[pair[0]] = pair[1]
+			if pair[0] == "targetGroupArn" {
+				targetGroupArn = true
+			}
+		}
+		containerPort, err := strconv.ParseInt(m["containerPort"], 10, 64)
```
iamhopaul123, line 127: Would be better to use `ParseUint` since port number >= 0.

iamhopaul123:
PS can you add unit test for this new change pls ❤️

ecs-cli/modules/cli/compose/entity/service/service.go
seongm-1: Yeah I like that idea a lot better!

ecs-cli/modules/commands/compose/service/compose_service_command.go
seongm-1: So I just put [deprecated] before the [Optional]?

ecs-cli/modules/commands/flags/flags.go
seongm-1: Yes I think that makes more sense :)

ecs-cli/modules/utils/utils.go
seongm-1: Oh that is a good idea!

ecs-cli/modules/utils/utils.go
seongm-1: I tried to convert it as a ParseUint however in the api.go file, it is listed as a int64. I tried to change it to a uint64 in that api.go but it warns me to not edit this file as it is generated so I didn't. We can keep it to a int64 unless you have other ideas?

ecs-cli/modules/commands/compose/service/compose_service_command.go
iamhopaul123:
```
		cli.StringFlag{
			Name: flags.ComposeServiceNamePrefixFlag,
			Usage: fmt.Sprintf(
				"[Deprecated] Specifies the prefix added to an ECS service created from a compose file. Format <prefix><project-name>. (defaults to empty)",
			),
		},
```

Maybe [Deprecated] instead of [Optional], since I think deprecated field cannot be necessary.

ecs-cli/modules/utils/utils.go
iamhopaul123: oh ok then just deep it to int64 :)

ecs-cli/modules/commands/compose/service/compose_service_command.go
seongm-1: Ah I just sent another PR out. Will do when include the change when I also add the unit tests

ecs-cli/modules/utils/utils.go
seongm-1: Oops, will add this next PR

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -117,20 +119,31 @@ func (s *Service) LoadContext() error {
 	// specified do not exist.
 	// TODO: Add validation on targetGroupArn or loadBalancerName being
 	// present if containerName or containerPort are specified
-	if role != "" || targetGroupArn != "" || loadBalancerName != "" || containerName != "" || containerPort != nil {
+	if role != "" || targetGroupArn != "" || loadBalancerName != "" || containerName != "" || containerPort != nil || len(targetGroupsList) != 0 {
 		if targetGroupArn != "" && loadBalancerName != "" {
 			return errors.Errorf("[--%s] and [--%s] flags cannot both be specified", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
+		} else if (targetGroupArn != "" || loadBalancerName != "") && len(targetGroupsList) != 0 {
+			return errors.Errorf("[--%s] or [--%s] flags cannot be specified with %s", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag, flags.TargetGroupsFlag)
```
iamhopaul123, line 126: it seems like we don't need to do this actually. I think it is ok to have both old flags and the new flag, since we can just concatenate them.

according to this doc https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-compose-service-up.html. we should have the following logic.
```suggestion
if targetGroupArn != "" || loadBalancerName != "" || containerName != "" || containerPort != nil {
  if role == "" {
    return error (role is needed)
  }
  if containerName == "" {
    return error (container name is required)
  }
  if containerPort == nil {
    return error
  }
  if targetGroupArn != "" && loadBalancerName != "" {
    return error
  }
  if targetGroupArn == "" && loadBalancerName == "" {
    return error
  }
  do sth
}
if len(targetGroupsList) != 0 {
  if role == "" {
    return error
  }
  do sth
}
```

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -132,8 +145,23 @@ func (s *Service) LoadContext() error {
...
+
+	if len(targetGroupsList) != 0 {
+		if role == "" {
+			return errors.Errorf("[--%s] is required when either when [--%s] flag is specified", flags.RoleFlag, flags.TargetGroupsFlag)
+		}
+		loadBalancers, err := utils.ParseLoadBalancers(targetGroupsList)
+		if err != nil {
+			return err
+		}
+		s.loadBalancers = append(s.loadBalancers, loadBalancers...)
```
iamhopaul123, line 161: Will this line fail if only new flag is specified? Since `s.loadBalancers` is not initialized yet.

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -132,8 +145,23 @@ func (s *Service) LoadContext() error {
 		if loadBalancerName != "" {
 			s.loadBalancer.LoadBalancerName = aws.String(loadBalancerName)
 		}
+		s.loadBalancers = []*ecs.LoadBalancer{s.loadBalancer}
+		// s.loadBalancers = append(s.loadBalancers, s.loadBalancer)
```
iamhopaul123, line 149: Maybe get rid of this?

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -45,6 +45,7 @@ type Service struct {
 	ecsContext        *context.ECSContext
 	deploymentConfig  *ecs.DeploymentConfiguration
 	loadBalancer      *ecs.LoadBalancer
```
iamhopaul123, line 47: I guess we don't need this variable now.

ecs-cli/modules/cli/compose/entity/service/service.go
seongm-1: Actually no, it is already initialized as an empty array

ecs-cli/modules/cli/compose/entity/service/service.go
seongm-1: When a user wants to add one load balancers, we will need this to put into an array because the createServiceInput currently accepts an array of load balancers.

ecs-cli/modules/cli/compose/entity/service/service.go
seongm-1: See above to why we might need this.

ecs-cli/modules/cli/compose/entity/service/service.go
iamhopaul123: I mean the line that gets commented out

ecs-cli/modules/cli/compose/entity/service/service.go
iamhopaul123: yeah but that init might get skipped if the old flags are not specified

ecs-cli/modules/cli/compose/entity/service/service.go
```diff
@@ -132,8 +145,23 @@ func (s *Service) LoadContext() error {
 		if loadBalancerName != "" {
 			s.loadBalancer.LoadBalancerName = aws.String(loadBalancerName)
 		}
+		s.loadBalancers = []*ecs.LoadBalancer{s.loadBalancer}
```
iamhopaul123, line 148: Yeah but it not worthwhile for defining another member variable for `Service`. Instead, it might be better to

```
loadBalancers = []*ecs.LoadBalancer{
   ...
}
....
s.loadBalancers = []*ecs.LoadBalancer{loadBalancer}
```

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,45 @@ func GetPartition(region string) string {
...
+			if pair[0] == "targetGroupArn" {
+				elbv2 = true
+			}
+		}
+		containerPort, err := strconv.ParseInt(m["containerPort"], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("Fail to parse container port %s for container %s", m["containerPort"], m["containerName"])
+		}
+		if elbv2 {
+			list = append(list, &ecs.LoadBalancer{
```
iamhopaul123, line 137: Might also need to do the validation as well to make sure
1) at least one of `targetGroupArn` and `loadBalancerName` exists
2) containerName and containerPort must exist

ecs-cli/modules/cli/compose/entity/service/service.go
seongm-1: So if only flags that are specified are the --role and --target-groups, it will not throw an error. Tbh I do not know the structure well enough to when it initializes it

ecs-cli/modules/utils/utils.go
```diff
@@ -116,6 +116,7 @@ func ParseLoadBalancers(flagValues []string) ([]*ecs.LoadBalancer, error) {
 	for _, flagValue := range flagValues {
 		m := make(map[string]string)
 		elbv2 := false
```
iamhopaul123, line 118: Just a nit but you can init like this: ` var elbv1, elbv2 bool`

ecs-cli/modules/utils/utils.go
```diff
@@ -139,12 +142,14 @@ func ParseLoadBalancers(flagValues []string) ([]*ecs.LoadBalancer, error) {
...
 			})
-		} else {
+		} else if elbv1 {
 			list = append(list, &ecs.LoadBalancer{
 				LoadBalancerName: aws.String(m["loadBalancerName"]),
 				ContainerName:    aws.String(m["containerName"]),
 				ContainerPort:    aws.Int64((containerPort)),
 			})
+		} else {
+			return nil, fmt.Errorf("InvalidParameterException: Target Group Arn or Load Balancer Name can not be blank")
```
iamhopaul123, line 152: Just `Target Group Arn or Load Balancer Name can not be blank`. "InvalidParameterException" is the error code defined by aws sdk go.

ecs-cli/modules/utils/utils.go
```diff
@@ -139,12 +142,14 @@ func ParseLoadBalancers(flagValues []string) ([]*ecs.LoadBalancer, error) {
 				ContainerName:  aws.String(m["containerName"]),
 				ContainerPort:  aws.Int64((containerPort)),
 			})
-		} else {
+		} else if elbv1 {
```
iamhopaul123, line 145: also need to throw an error when both elbv1 and elbv2 are true.

ecs-cli/modules/utils/utils.go
```diff
@@ -131,6 +131,9 @@ func ParseLoadBalancers(flagValues []string) ([]*ecs.LoadBalancer, error) {
 			} else if pair[0] == "loadBalancerName" {
 				elbv1 = true
 			}
+			if elbv1 && elbv2 {
+				return nil, errors.Errorf("[--%s] and [--%s] flags cannot both be specified", "target-group-arn", "load-balancer-name")
```
iamhopaul123, line 135: nit: fmt.Errorf

ecs-cli/modules/cli/compose/entity/service/service_test.go
```diff
@@ -778,7 +1018,7 @@ func getCreateServiceWithDelayMockClient(t *testing.T,
 		DesiredCount:   aws.Int64(1),
 		RunningCount:   aws.Int64(1),
 		ServiceName:    aws.String("test-created"),
-		Deployments:    []*ecs.Deployment{
+		Deployments: []*ecs.Deployment{
```
seongm-1, line 1021: I indent it but when I save the file, it automatically reverts. Any idea what is going on?

iamhopaul123: lol it's done by your IDE. Don't force it and let it be.

seongm-1: I just have no idea what is causing it. Nothing much I can do about it lol

bvtujo: more specifically it's probably GoFmt deciding something arcane about the indentation of struct literal members.

iamhopaul123: Or if it is not consistent to the other ones. You can save it using other text editors.

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,68 @@ func GetPartition(region string) string {
...
+		}
+		var elbv1, elbv2 bool
+		keyValPairs := strings.Split(flagValue, ",")
+
+		for _, kv := range keyValPairs {
+			pair := strings.SplitN(kv, "=", -1)
+			if len(pair) > 2 {
+				return nil, fmt.Errorf("Only include one = to indicate your value in your %s", pair[0])
+			}
+			if pair[0] != "targetGroupArn" && pair[0] != "loadBalancerName" && pair[0] != "containerName" && pair[0] != "containerPort" {
```
iamhopaul123, line 130: Would be better to define key, val := pair[0], pair[1] to make it clearer. Also can you make "targetGroupArn"... those to be constant vars?

ecs-cli/modules/utils/utils.go
```diff
@@ -106,3 +107,68 @@ func GetPartition(region string) string {
...
+				ContainerPort:  aws.Int64((containerPort)),
+			})
+		} else if elbv1 {
+			list = append(list, &ecs.LoadBalancer{
+				LoadBalancerName: aws.String(m["loadBalancerName"]),
+				ContainerName:    aws.String(m["containerName"]),
+				ContainerPort:    aws.Int64((containerPort)),
+			})
+		} else {
+			return nil, fmt.Errorf("Target Group Arn or Load Balancer Name can not be blank")
```
iamhopaul123, line 170: nit: cannot

ecs-cli/modules/cli/compose/entity/service/service_test.go
```diff
@@ -362,6 +362,281 @@ func TestCreateWithHealthCheckGracePeriodAndALB(t *testing.T) {
 	)
 }
 
+func TestCreateWithTwoTargetGroupsFlag(t *testing.T) {
+	role := "role"
+	targetGroups := &cli.StringSlice{}
+	targetGroups.Set("targetGroupArn=CS:GO,containerName=Bobs,containerPort=80")
+	targetGroups.Set("targetGroupArn=Overwatch,containerName=Joes,containerPort=40")
+
+	targetGroupArnList := []string{"CS:GO", "Overwatch"}
```
iamhopaul123, line 371: lol

ecs-cli/modules/commands/compose/service/compose_service_command.go
```diff
@@ -222,10 +222,11 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
-	containerNameUsageString := fmt.Sprintf("[Optional] Specifies the container name (as it appears in a container definition). This parameter is required if --%s or --%s is specified.", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
-	containerPortUsageString := fmt.Sprintf("[Optional] Specifies the port on the container to associate with the load balancer. This port must correspond to a containerPort in the service's task definition. This parameter is required if --%s or --%s is specified.", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
-	loadBalancerNameUsageString := fmt.Sprintf("[Optional] Specifies the name of a previously configured Classic Elastic Load Balancing load balancer to associate with your service. NOTE: For Application Load Balancers or Network Load Balancers, use the --%s flag.", flags.TargetGroupArnFlag)
+	targetGroupArnUsageString := fmt.Sprintf("[Deprecated] Specifies the full Amazon Resource Name (ARN) of a previously configured target group for an Application Load Balancer or Network Load Balancer to associate with your service. NOTE: For Classic Load Balancers, use the --%s flag.", flags.LoadBalancerNameFlag)
```
efekarakus, line 225: Although these are marked as [DEPRECATED] they will still work right, there is no regression for the customers?

ecs-cli/modules/commands/compose/service/compose_service_command.go
```diff
@@ -222,10 +222,11 @@ func deploymentConfigFlags(specifyDefaults bool) []cli.Flag {
...
+	containerNameUsageString := fmt.Sprintf("[Deprecated] Specifies the container name (as it appears in a container definition). This parameter is required if --%s or --%s is specified.", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
+	containerPortUsageString := fmt.Sprintf("[Deprecated] Specifies the port on the container to associate with the load balancer. This port must correspond to a containerPort in the service's task definition. This parameter is required if --%s or --%s is specified.", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
+	loadBalancerNameUsageString := fmt.Sprintf("[Deprecated] Specifies the name of a previously configured Classic Elastic Load Balancing load balancer to associate with your service. NOTE: For Application Load Balancers or Network Load Balancers, use the --%s flag.", flags.TargetGroupArnFlag)
+	targetGroupsUsageString := fmt.Sprintf("[Optional] Specifies multiple target groups to register with a service. Cannot be used with --%s flag or --%s at the same time.", flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
```
efekarakus, line 229: What do you think about including an example to show how this flag can be used? It seems like the format is not super obvious for customers.

```suggestion
	targetGroupsUsageString := fmt.Sprintf(`[Optional] Specifies multiple target groups to register with a service. Cannot be used with --%s flag or --%s at the same time.
Example: --target-groups "targetGroupArn=arn,containerName=nginx,containerPort=80"`, flags.LoadBalancerNameFlag, flags.TargetGroupArnFlag)
```

ecs-cli/modules/commands/compose/service/compose_service_command.go
seongm-1: Correct, the user can still use these flags to register a single load balancer

ecs-cli/modules/commands/compose/service/compose_service_command.go
seongm-1: That sounds a like a good idea!

ecs-cli/modules/utils/utils.go
```diff
@@ -130,6 +135,7 @@ func ParseLoadBalancers(flagValues []string) ([]*ecs.LoadBalancer, error) {
 			if len(pair) > 2 {
```
iamhopaul123, line 135: Would it be better to have like this
```suggestion
			if len(pair) != 2 {
			  return nil, fmt.Errorf("Only include one = to indicate your value in your %s", key)
			}
			key, val := pair[0], pair[1]
```
so that it is able to catch the case when input does not have any "=" and prevent the golang panic.

ecs-cli/modules/utils/utils.go
```diff
@@ -27,7 +27,11 @@ import (
 
 const (
 	// ECSCLIResourcePrefix is prepended to the names of resources created through the ecs-cli
-	ECSCLIResourcePrefix = "amazon-ecs-cli-setup-"
+	ECSCLIResourcePrefix     = "amazon-ecs-cli-setup-"
+	containerNameParamKey    = "containerName"
```
iamhopaul123, line 31: :+1:

ecs-cli/modules/utils/utils.go
seongm-1: I like this! One issue is where the key is not initialized before. So I could either do pair[0] for the error message or not include the %s and just display an error message.

PR #1031, opened 2020-05-22
Update buildspecs to work with dep & codebuild standard 4.0

bvtujo:
Buildspec.yml now copies all files to the gopath and works from there, not the codebuild source directory. 

Buildspec_integ.yml does not set the GOPATH, GOROOT, or GOCACHE before building the test binary. This was causing a conflict with goenv and the shim-based go versioning used in standard images >=2.0. 
Prerequisite work for version 1.19.0

This change introduces a failure (only in Codebuild on AWS) of one unit test, perhaps related to an environment variable being set unexpectedly in the build image.

```
607 | --- FAIL: TestNewCommandConfigFromEnvVarsWithRegionNotSpecified (0.00s)
--
608 | command_config_test.go:86:
609 | Error Trace:    command_config_test.go:86
610 | Error:          An error is expected but got nil.
611 | Test:           TestNewCommandConfigFromEnvVarsWithRegionNotSpecified
612 | Messages:       Expected error when region is not specified
```

----

**Testing**
- [x ] Unit tests passed
- [ x] Integration tests passed
- [n/a ] Unit tests added for new functionality

----

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -44,6 +44,12 @@ phases:
       - CHECKED_IN_VERSION=`cat /go/src/github.com/aws/amazon-ecs-cli/VERSION`
       - echo "VERSION_FILE=$CHECKED_IN_VERSION"
       - echo "GOPATH=$GOPATH"
+      # Check if there is a tag currently pointing at the head and exit if not. 
+      - |
+        if [ -z "$VERSION" ]; then
+          echo "the current commit is not tagged. Stopping the build stage..."
+          exit 1
+        fi
       - |
         # $CHECKED_IN_VERSION is not prefixed with "v", only the semantic version number,
         # such as 1.17.0 instead of v1.17.0, which is what normal version tags look like.
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -6,10 +6,11 @@ phases:
       golang: 1.13
   pre_build:
     commands:
-      # GOPATH is setup like the following in Codebuild standard 2.0 image
+      # GOPATH is setup like the following in Codebuild standard 4.0 image
       # /go:/codebuild/output/src<some numbers>
-      # so we copy all the source code to the appropriate location before
-      # invoking any go command
+      # so we copy all the source code to the go root install location before
+      # invoking any go command. 
+      # this also allows us to build using dep while the transition to go modules is ongoing
       - ls -lah
       - mkdir -p /go/src/github.com/aws/amazon-ecs-cli
       - cp -R . /go/src/github.com/aws/amazon-ecs-cli/
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -32,7 +32,7 @@ phases:
       - GOOS=linux GOARCH=amd64 ECS_RELEASE=cleanbuild go run gen/version-gen.go
   build:
     commands:
-      - echo "cd into $CODEBUILD_SRC_DIR"
+      - echo "cd into go root directory"
       - cd /go/src/github.com/aws/amazon-ecs-cli
       - echo "Compilation context:"
       - echo "CODEBUILD_SOURCE_VERSION=$CODEBUILD_SOURCE_VERSION"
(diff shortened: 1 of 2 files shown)


buildspec_integ.yml
```diff
@@ -9,7 +8,11 @@ phases:
       - mkdir -p /go/src/github.com/aws/amazon-ecs-cli
       - cp -R . /go/src/github.com/aws/amazon-ecs-cli/
       - cd /go/src/github.com/aws/amazon-ecs-cli/
+      - echo "installing dependencies..."
+      - go get github.com/wadey/gocovmerge
```
allisaurus, line 12: why do we need to add coverage reporting to this stage? I'm not against it in principle but it will add time/dependencies  and we have it running in other build projects

allisaurus:
btw is this accurate anymore? (should update if not) 
https://github.com/aws/amazon-ecs-cli/blob/master/buildspec.yml#L9

buildspec.yml
```diff
@@ -39,16 +44,22 @@ phases:
       - CHECKED_IN_VERSION=`cat /go/src/github.com/aws/amazon-ecs-cli/VERSION`
       - echo "VERSION_FILE=$CHECKED_IN_VERSION"
       - echo "GOPATH=$GOPATH"
+      # Check if there is a tag currently pointing at the head and exit if not. 
+      - |
+        if [ -z "$VERSION" ]; then
+          echo "the current commit is not tagged. Stopping the build stage..."
+          exit 1
+        fi
```
hencrice, line 52: Did we miss this previously? This seems like a generic validation improvement

buildspec.yml
```diff
@@ -6,29 +6,34 @@ phases:
...
+      # so we copy all the source code to the go root install location before
+      # invoking any go command. 
+      # this also allows us to build using dep while the transition to go modules is ongoing
       - ls -lah
       - mkdir -p /go/src/github.com/aws/amazon-ecs-cli
       - cp -R . /go/src/github.com/aws/amazon-ecs-cli/
-      - |
-        env -i PATH=$PATH GOPATH=`go env GOPATH` GOROOT=`go env GOROOT` GOCACHE=`go env GOCACHE` \
-        go test -race -v -cover github.com/aws/amazon-ecs-cli/ecs-cli/modules/...
+      - cd /go/src/github.com/aws/amazon-ecs-cli
```
hencrice, line 17: In general, I avoid `cd` into any specific directory because we might forget to cd out before executing later commands. But I don't have a strong preference on this.

buildspec.yml
```diff
@@ -6,29 +6,34 @@ phases:
...
       # need to cd into the version package because version-gen.go assumes the relative
       # location of the VERSION file.
-      - cd /go/src/github.com/aws/amazon-ecs-cli/ecs-cli/modules/version/
+      - cd ./ecs-cli/modules/version/
       # since we are running the go program inside a Linux container, has to hardcode
       # the GOOS and GOARCH correspondinly regardless of what the host OS is.
       - GOOS=linux GOARCH=amd64 ECS_RELEASE=cleanbuild go run gen/version-gen.go
   build:
     commands:
       - echo "cd into $CODEBUILD_SRC_DIR"
```
allisaurus, line 35: this msg doesn't seem to be accurate anymore

buildspec.yml
bvtujo: Yes, it seems like we thought that the check against the checked in version would cover this case, but the pipeline stage was succeeding when there was no git tag. I thought this would be a good time to add that validation to make sure it behaves the way we thought it did.

buildspec_integ.yml
```diff
@@ -9,7 +8,11 @@ phases:
...
+      - echo "building test binary..."
+      - go test -coverpkg ./ecs-cli/modules/... -c -tags testrunmain -o ./bin/local/ecs-cli.test ./ecs-cli
```
allisaurus, line 14: would it be better to build the binary in the `build` stage as before, or is there some context here re: env vars and dirs that makes it better to execute here? (not sure if shell context changes between build phases)

bvtujo: exported variables get cleared between phases but files persist. Because this generates a file it doesn't particularly matter where it happens. Happy to move it for consistency.

